PR #3, opened 2016-11-02
Add function for document creation

aenario:


jinroh:
I think its missing a return [in the `promiser ` function](https://github.com/cozy/cozy-api/blob/73775d872a8681b81c08da3796ef0cb6e67c383b/src/utils.js#L16) to make it chainable.

aenario:
I will include the promiser return in my next PR for clarity, but it will be a return null : 

promiser goal is to exit Promise-dreamland to fall back into callback-hell if the developper is so inclined. It should always be the last call in a promise chain.

aenario:
I will include the promiser return in my next PR for clarity, but it will be a return null : 

promiser goal is to exit Promise-dreamland to fall back into callback-hell if the developper is so inclined. It should always be the last call in a promise chain.

aenario:
I will include the promiser return in my next PR for clarity, but it will be a return null : 

promiser goal is to exit Promise-dreamland to fall back into callback-hell if the developper is so inclined. It should always be the last call in a promise chain.

aenario:
I will include the promiser return in my next PR for clarity, but it will be a return null : 

promiser goal is to exit Promise-dreamland to fall back into callback-hell if the developper is so inclined. It should always be the last call in a promise chain.

aenario:
I will include the promiser return in my next PR for clarity, but it will be a return null : 

promiser goal is to exit Promise-dreamland to fall back into callback-hell if the developper is so inclined. It should always be the last call in a promise chain.

aenario:
I will include the promiser return in my next PR for clarity, but it will be a return null : 

promiser goal is to exit Promise-dreamland to fall back into callback-hell if the developper is so inclined. It should always be the last call in a promise chain.

aenario:
I will include the promiser return in my next PR for clarity, but it will be a return null : 

promiser goal is to exit Promise-dreamland to fall back into callback-hell if the developper is so inclined. It should always be the last call in a promise chain.

jinroh:
I see ! LGTM then.

PR #4, opened 2016-11-02
Add integration tests against cozy-stack in travis

aenario:


.travis.yml
```diff
@@ -1,3 +1,29 @@
...
+  - docker run -d -p 5984:5984 --name couch klaemo/couchdb:2.0.0
+  - sleep 10
+  - curl -X PUT http://127.0.0.1:5984/{_users,_replicator,_global_changes}
+  # go 1.7
+  - eval "$(gimme 1.7)"
+  # start a stack instance from master
+  - mkdir $GOPATH
+  - go get -u github.com/cozy/cozy-stack
+  - go run $GOPATH/src/github.com/cozy/cozy-stack/main.go instances add dev
+  - go run $GOPATH/src/github.com/cozy/cozy-stack/main.go serve &
```
nono: I think you can replace `go run $GOPATH/src/github.com/cozy/cozy-stack/main.go` with just `$GOPATH/bin/cozy-stack`

.travis.yml
```diff
@@ -1,3 +1,29 @@
...
+services:
+  - docker
+
+env:
+  GOPATH: ${TRAVIS_BUILD_DIR}/_workspace
+
+hosts:
+    - alice.cozy.io
+    - bob.cozy.io
+    - charlie.cozy.io
```
nono: By curiosity, why **3** hosts?

nono: And I think it would be safer to use something like `alice.cozy.local`. It avoids to ping a real cozy instance the day we forget to declare here a fouth host used in tests.

aenario: The hosts are not actually used for now, but I figured we will need them later to tests CORS and auth (with karma) and didn't want to forget how it is done in travis.

PR #6, opened 2016-11-09
Add update and delete methods

jinroh:


nono:
@jinroh did you see that travis build is red?

jinroh:
@nono fixing it (test:v1)

src/crud.js
```diff
@@ -35,14 +60,66 @@ export async function create (doctype, attributes) {
...
+
+  let path = createPath(config, doctype, _id)
+  let response = await doFetch(config, 'PUT', path, changes)
+
+  if (config.isV1) return await find(doctype, _id)
+
+  return response.data
+}
+
+export async function _delete (doctype, doc) {
```
aenario: why the _ ?

src/crud.js
```diff
@@ -35,14 +60,66 @@ export async function create (doctype, attributes) {
...
+export async function update (doctype, doc, changes) {
+  const config = await waitConfig()
+
+  const {_id, _rev} = doc
+
+  if (!_id) {
+    throw new Error('Missing _id field in passed document')
+  }
+
+  if (!_rev) {
```
aenario: _rev is only mandatory for stack v2

src/crud.js
jinroh: because babel (or one tool of the build chain) would not allow it, throwing a "Unexpected token".. 😞 . if you have a way to get around this, i'm interested.

src/crud.js
jinroh: indeed.. e6c5530937e4e56258ab611b962d45d8cc02d035

src/crud.js
nono: `delete` is a reserved keyword. You can't use it as a function name

PR #8, opened 2016-11-23
Mango and misc

aenario:
This still needs more tests around edge cases, but opening a PR for @jinroh to review.

### DIFF
--- a/src/mango.js
+++ b/src/mango.js
@@ -1,0 +1,232 @@
+import {warn, waitConfig, createPath, doFetch, normalizeDoctype} from './utils'
+
+export async function defineIndex (doctype, fields) {
+  const config = await waitConfig()
+  doctype = normalizeDoctype(config, doctype)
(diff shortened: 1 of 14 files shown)


src/mango.js
```diff
@@ -0,0 +1,232 @@
...
+  return response.docs
+}
+
+// misc
+function capitalize (name) {
+  return name.charAt(0).toUpperCase() + name.slice(1)
+}
+
+function makeMapFunction (doctype, fields) {
+  if (Object.prototype.toString.call(fields) !== '[object Array]' ||
```
jinroh: A reason not to use `Array.isArray` ?

aenario: mostly FUD i guess :-)

aenario: fixed in https://github.com/cozy/cozy-client-js/pull/9/commits/b4712a416c955b7c066da92d0626345e3c1ae1b1 which is part of #9

PR #14, opened 2016-11-28
upload

jinroh:
This PR adds an `upload(data, options)` function. This upload function can be used in `create` or `update` mode. See the documentation for more details on the way this function works.

-----

NOTE: I tried to unify a little bit better the documentation and tried to follow this template:

```
### cozy.function(arg1, arg2)

`cozy.function(arg1, arg2)` is used to ....

It returns ...

If it ... (error cases)

**Warning** (if any)

- arg1 is a string specifying ...
- arg2 is an object with the following fields
  * field1 ...
  * field2 ...

example...
```

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -283,7 +283,7 @@ const created = await cozy.createFile(blob, {
     name: "filename",
     folderId: "123456",
 })
-const fileUpdated = await cozy.createFile(fileInput.files[0], { folderId: "" })
+const fileCreated = await cozy.createFile(fileInput.files[0], { folderId: "" })
 ```
 
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,5 +1,5 @@
 [Cozy](https://cozy.io) Javascript Client
-==============================
+=========================================
 
 `cozy-client-js` is a javascript library made by Cozy. It enables client-side  applications to make requests to the cozy stack.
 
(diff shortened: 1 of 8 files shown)


aenario:
How about having two separate functions ?

- vfs.createFile(parentFolderID, name, contentBlob)
- vfs.updateContent(FileID, contentBlob)

Where contentBlob is either
- a File (with content-type included)
- an object {contentType:"xxx", data: Blob|ArrayBuffer|ArrayBufferView|string}

Otherwise, everything LGTM and :+1: for doc cleanup

jinroh:
I agree with you for the separation in two functions.

However, I would suggest something like that:

- `cozy.createFile({ name, folderId, data, contentType })`
- `cozy.updateFile({ fileId, data, contentType })`

or 

- `cozy.createFile(data, { name, folderId, contentType })`
- `cozy.updateFile(data, { fileId, contentType })`

I think it would allow us to add more parameters to have an "options" object... For example, to allow creating a file by its path. What do you think ?

aenario:
I may be crazy but I cant find where you pass the data to fetch... Is the file created not empty?

jinroh:
@aenario you're not crazy ! the files were empty. i fixed the issue.

i'll close this PR for #15 which contains upload, mkdir and trash methods.

PR #15, opened 2016-11-28
createFile, updateFile, createDirectory and trash

jinroh:
Add a `createFile`, `updateFile`, `createDirectory` and a `trash` function. ~~To merge after #14.~~

### DIFF
--- a/test/integration/files.js
+++ b/test/integration/files.js
@@ -8,71 +8,48 @@ global.fetch = fetch
 
 const TARGET = process.env && process.env.TARGET || ''
 
-describe('files API', function () {
+describe('files API', async function () {
   let config
 
-  describe('Config', function () {
-    it('When called against a real instance', async function () {
-      config = await cozy.init({ target: TARGET })
-    })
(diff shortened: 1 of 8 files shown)


test/integration/files.js
```diff
@@ -47,4 +47,17 @@ describe('files API', function () {
       updated.data.should.have.property('attributes')
     })
   })
+
+  describe('Create directory', function () {
+    it('Works', async function () {
+      if (config.isV2) {
```
nono: Can we move this check a bit upper? So, we can avoid to have `Create directory Works` in the logs when running the integration tests with cozy V2.

jinroh:
I'm wondering whether or not we should rename the `updateFile` to `updateFileContent` to keep `updateFile` for updating the file's metadata. Either that or we have to extend the possibilities of the `updateFile` method to support both (re-upload + change metadata).

aenario:
Merging as-is so I can rebase on it. We can rename in a separate PR, let's discuss it on daily.

PR #23, opened 2016-12-22
Add stateless methods to perform OAuth

jinroh:


docs/README.md
```diff
@@ -399,6 +399,137 @@ const buff = await response.arrayBuffer()
...
+### `cozy.auth.Token`
+
+`cozy.auth.Token` is a class representing an OAuth token.
+
+```
+type Token {
+  tokenType: string;
+  accessToken: string;
+  refreshToken: string;
+  scope: string
```
nono: I'm a bit puzzled here. What are `tokenType` and `scope`? And there should be a registrationToken (or registrationAccessToken) here, no?

src/auth.js
```diff
@@ -0,0 +1,223 @@
...
+    client = new Client(client.url, client)
+  }
+  if (!client.isRegistered()) {
+    return Promise.reject(new Error('Client not registered'))
+  }
+  return fetch(`${client.url}/auth/register/${client.clientID}`, {
+    method: 'GET',
+    headers: {
+      'Accept': 'application/json',
+      'Authorization': token.toAuthHeader()
```
nono: Nope, it's not the access_token that is used here, but the registration_access_token!

docs/README.md
```diff
@@ -515,9 +513,9 @@ new Client(url, options)
 ```
 
 
-### `cozy.auth.Token`
+### `cozy.auth.AccessToken`
 
-`cozy.auth.Token` is a class representing an OAuth token.
+`cozy.auth.AcessToken` is a class representing an OAuth access token.
```
nono: s/Acess/Access/

docs/README.md
```diff
@@ -450,15 +448,15 @@ const {url, state} = cozy.auth.getAuthCodeURL(client, ['files/images:read'])
 
 // save state and redirect to url
 localStorage.setItem("oauthstate", state)
-document.location.replace(url)
+window.location.replace(url)
```
nono: I think there are over places where `document.location` is used and should be replaced with `window.location`

docs/README.md
```diff
@@ -399,6 +399,135 @@ const buff = await response.arrayBuffer()
...
+
+- `url` is a string of the url of the cozy
+- `options` is an object with the same fields as a client object, or is an instance of client
+
+```
+new Client(url, options)
+```
+
+
+### `cozy.auth.AccessToken`
```
nono: It is still cozy.auth.Token in the code

PR #24, opened 2016-12-27
Add stateful OAuth flow

jinroh:
**Still WIP**

This PR adds an stateful authentication management for both V2 and V3 stack (with OAuth).

-------


src/auth_storage.js
```diff
@@ -0,0 +1,98 @@
+export class LocalStorage {
+  constructor (prefix = 'cozy:oauth:', getLocalStorage = () => window.localStorage) {
```
nono: I don't see why you pass a function as last parameter, instead of just the selected storage

src/auth_storage.js
```diff
@@ -0,0 +1,98 @@
...
+    // We try to deref the localstorage, and if it fails (because the user has
+    // deactivated its localstorage for instance) we fallback on a memory
+    // storage.
```
nono: I don't think it's a good idea to silently fallback to memstorage if localStorage can't be used. The typical use case for localStorage is a SPA, and memStorage won't work in that case because the page will be unloaded and reloaded later.

src/auth_storage.js
```diff
@@ -0,0 +1,98 @@
...
+  save (key, value) {
+    return new Promise(resolve => resolve(
+      this.storage.setItem(this.prefix + key, JSON.stringify(value))))
+  }
+
+  load (key) {
+    return new Promise(resolve => {
+      const item = this.storage.getItem(this.prefix + key)
+      if (!item) {
+        resolve()
```
nono: If item is not found, shouldn't be it a `reject` instead of a `resolve`?

src/auth_storage.js
```diff
@@ -0,0 +1,98 @@
...
+        }
+      }
+      resolve()
+    })
+  }
+}
+
+export class MemoryStorage {
+  constructor () {
+    this.hash = {}
```
nono: You should probably use an ES2015 Map, or `Object.create(null)`, instead of just `{}`

nono: You can also use `hasOwnProperty` instead of `[]`

src/auth_v3.js
```diff
@@ -148,53 +135,139 @@ export function getAuthCodeURL (client, scopes = []) {
...
+export function refreshToken (cozy, client, token) {
+  return retrieveToken(cozy, client, token, {
     'grant_type': 'refresh_token',
     'code': token.refreshToken
   })
 }
 
+// oauthFlow perform the stateful registration and access granting of an OAuth
+// client.
+export function oauthFlow (cozy, storage, pageURL, createClient, onRegistered) {
```
nono: It's hard to read this function. Maybe using `await` can help to make it more straightforward

webpack.config.js
```diff
@@ -4,11 +4,13 @@ var nodeExternals = require('webpack-node-externals')
 var path = require('path')
 var webpack = require('webpack')
 
+var NODE_ENV = process.env.NODE_ENV
+var NODE_TARGET = process.env.NODE_TARGET || 'web'
+
 var config = {
-  entry: ['whatwg-fetch', 'regenerator-runtime/runtime', path.join(__dirname, 'src', 'index.js')],
```
nono: You have removed `async/await` support?

PR #28, opened 2017-01-02
Improve API and doc and add complete working example with node.js

jinroh:
This PR should improve the documentation around the OAuth usage of the library.

It also adds an export of the library for nodejs as well as a complete example of how to use the library for an OAuth client with nodejs.

By implementing this example client I made a change in the defined API so that the `onRegistered` callback is now expected to return a Promise of the URL that the client should request containing the code and state. See the example for more details.

Also, the `oauthFlow` method should hopefully be a little simpler, and the options of OAuth passed to the constructor or init method are now in a separate object field named `oauth`.


jinroh:
Rebased on master

PR #30, opened 2017-01-05
Default to cookie-base authentication if no oauth parameters are given

jinroh:
By implementing OAuth, I removed the possibility to use cookie-based authentification.

This PR makes it the default (ie. not using `Authorization` header) if no `oauth` parameters are given to the `init` method.


aenario:
@jinroh To send cookies in modern browser (which have a native fetch function), you need to explicitly enable it by adding a credentials option to fetch (ref https://github.com/github/fetch#sending-cookies)

PR #38, opened 2017-01-18
Add offline database :airplane:

kosssi:
This feature allow to create pouch database on client side to facilitate offline first.

- [x] Create databases on init for each docType you want and pouchdb options
- [x] Create database after init for one docType and pouchdb options
- [x] Add unit test
- [x] Build library
- [ ] Update documentation


src/offline.js
```diff
@@ -0,0 +1,16 @@
+const PouchDB = require('pouchdb')
+
+export function init (cozy, offline) {
```
enguerran: We could save 2 lines with ES2015 syntax, couldn´t we ?

```js
export function init (cozy, { options = {}, docTypes = [] }) {
  for (let docType of docTypes) {
    addDocType(cozy, docType, options)
  }
}
```

src/offline.js
```diff
@@ -0,0 +1,16 @@
...
+  const options = offline.options || {}
+  const docTypes = offline.docTypes || []
+  for (let docType of docTypes) {
+    addDocType(cozy, docType, options)
+  }
+}
+
+export function addDocType (cozy, docType, options) {
+  if (cozy._offline_databases === false) {
```
enguerran: We may check if it is an array, is there another way to check that like `if (!Array.isArray(cozy._offline_databases) { cozy._offline_databases = [] }`

nono: Or `cozy._offline_databases = cozy._offline_databases || []`

aenario:
Looks like there is some conflict in leveldown installation between cozy-data-system & pouchdb ... 

You can try : 
- using yarn to install both
- installing data-system & proxy with --global

Good luck :clinking_glasses:

package.json
```diff
@@ -61,5 +64,8 @@
     "All": {
       "description": "Test with any doctype."
     }
+  },
+  "dependencies": {
+    "pouchdb": "6.1.1"
```
enguerran: @kosssi could you explain why you had this dependency? I though we let the third-party application to add or not pouchdb in its dependencies.

webpack.config.js
```diff
@@ -45,6 +45,9 @@ var config = {
   resolve: {
     root: path.resolve('./src'),
     extensions: ['', '.js']
+  },
+  externals: {
+    pouchdb: 'PouchDB'
```
enguerran: @kosssi it may be a typo here, cf. https://github.com/cozy/cozy-client-js/pull/43

package.json
enguerran: Maybe it has to be in `devDependencies` if tests used it. Or in `peerDependencies` if not. I am not a pro with that.

PR #41, opened 2017-01-19
Add replication from Cozy :inbox_tray:

kosssi:


src/offline.js
```diff
@@ -6,7 +6,26 @@ export function init (cozy, { options = {}, docTypes = [] }) {
...
+      'change', 'paused', 'active', 'denied', 'complete', 'error'
+    ]
+    for (let eventName of eventNames) {
+      if (typeof events[eventName] === 'function') {
+        replication.on(eventName, events[eventName])
+      }
+    }
+    return replication
+  } else {
+    throw new Error(`You could add this docType: ${docType} to offline.`)
```
aenario: s/could/should/

src/offline.js
```diff
@@ -6,7 +6,26 @@ export function init (cozy, { options = {}, docTypes = [] }) {
...
+
+export function replicateFromCozy (cozy, docType, options = {}, events = {}) {
+  if (cozy._offline_databases && docType in cozy._offline_databases) {
+    const url = cozy._url + '/data/' + docType
+    let db = cozy._offline_databases[docType]
+    let replication = db.replicate.from(url, options)
+    const eventNames = [
+      'change', 'paused', 'active', 'denied', 'complete', 'error'
+    ]
+    for (let eventName of eventNames) {
```
aenario: if you return the replication, I am not sure we need the events: the using code can call .on(x) itself.

test/integration/offline.js
```diff
@@ -0,0 +1,52 @@
...
+      for (var i = 0, l = docs.length; i < l; i++) {
+        await cozy.delete(DOCTYPE, docs[i])
+      }
+    }
+  })
+
+  it('can replicate database from cozy', async function () {
+    cozy.offline.addDocType(DOCTYPE, {adapter: 'memory'})
+    let complete = await cozy.offline.replicateFromCozy(DOCTYPE)
+    complete.docs_written.should.not.equal(67)
```
aenario: ?

kosssi:
it's done @aenario ;)

src/offline.js
kosssi: it's for not touch replication and simplify it.

src/offline.js
```diff
@@ -1,12 +1,53 @@
...
+
+export function hasDatabase (cozy, doctype) {
+  return cozy._offline_databases && doctype in cozy._offline_databases
+}
+
+export function getDatabase (cozy, doctype) {
+  if (hasDatabase(cozy, doctype)) {
+    return cozy._offline_databases[doctype]
+  }
+  return false
```
enguerran: Why `getDatabase` returns a pouchDB instance (`Object`) or `false`. I know JavaScript allows weak typing, but shouldn't we return consistant JavaScript types like `Object || undefined`.

test/integration/offline.js
```diff
@@ -0,0 +1,65 @@
...
+    })
+    for (var i = 0, l = docs.length; i < l; i++) {
+      docs[i] = await cozy.create(DOCTYPE, docs[i])
+    }
+  })
+
+  after(async function () {
+    if (COZY_STACK_VERSION === '3') {
+      for (var i = 0, l = docs.length; i < l; i++) {
+        await cozy.delete(DOCTYPE, docs[i])
```
enguerran: I am curious about this, what the meaning of `l`?
Why don't use:
```javascript
docs.forEach(function(doc) { await cozy.delete(DOCTYPE, doc) })
```

test/integration/offline.js
```diff
@@ -0,0 +1,65 @@
...
+
+  before(async function () {
+    if (COZY_STACK_VERSION === '2') {
+      return this.skip()
+    }
+    cozy = new Cozy({
+      cozyURL: COZY_STACK_URL
+    })
+    for (var i = 0, l = docs.length; i < l; i++) {
+      docs[i] = await cozy.create(DOCTYPE, docs[i])
```
enguerran: same remark as [below](https://github.com/cozy/cozy-client-js/pull/41/files#r97268831)

```javascript
docs = docs.map(function(doc) { return await cozy.create(DOCTYPE, doc) })
```

test/integration/offline.js
```diff
@@ -0,0 +1,65 @@
...
+
+  after(async function () {
+    if (COZY_STACK_VERSION === '3') {
+      for (var i = 0, l = docs.length; i < l; i++) {
+        await cozy.delete(DOCTYPE, docs[i])
+      }
+    }
+  })
+
+  it('can replicate database from cozy', async function () {
```
enguerran: I cannot understand the intention beyond the 
```javascript
let complete = await cozy.offline.replicateFromCozy(DOCTYPE)
complete.docs_written.should.not.equal(0)
complete = await cozy.offline.replicateFromCozy(DOCTYPE)
complete.docs_written.should.equal(0)
```

test/integration/offline.js
aenario: Does not work with await, you'd have to use Promise.all, but i find it less readable

test/integration/offline.js
aenario: First replication does something, second does not (as the two databases are already sync)

test/integration/offline.js
kosssi: `l` is for lenght ;) I just copy the code from https://github.com/cozy/cozy-client-js/blob/master/test/integration/mango.js

kosssi: forEach is not compatible with await.

kosssi:
I think it ok ;)

test/integration/offline.js
enguerran: regarding [@aenario 's comment](https://github.com/cozy/cozy-client-js/pull/41#discussion_r97283316), it is not.

Why do you instantiate two variables in the for loop? I guess `for (var i = 0 ; i < docs.length ; i++)` should work the same as `for (var i = 0, l = docs.length ; i < l ; i++)`. Or do I misunderstand something?

cf. https://jsperf.com/array-prototype-length/1 ;)

test/integration/offline.js
enguerran: okay

test/integration/offline.js
aenario: Yep, it is a stupid reflex I got from somewhere, which probably got it from the time before JS runtimes optimised it. Good on kossi to follow the repo codestyle, but we can change it as I agree it is less readable.

PR #42, opened 2017-01-24
OAuth: fix the registration token storage

jinroh:
Since the `registration_access_token` is only given on registration, after a new `getClient`, the token was erased from the storage. This PR should fix that by passing the token from the old client generation to the next on `getClient` and `updateClient`. This way the token should be persisted correctly.

Unit tests were not mocking the API correctly. This PR also fixes that to test the new code path.


nono:
https://github.com/cozy/cozy-stack/pull/240

src/auth_v3.js
```diff
@@ -120,7 +120,25 @@ export function getClient (cozy, client) {
...
+// data of the client, from the API, and an old instance of the client.
+function createClient (data, oldClient) {
+  const newClient = new Client(data)
+  // we need to keep track of the registrationAccessToken since it is send
+  // only on registration. The GET /auth/register/:client-id endpoint does
+  // not return this token.
+  const shouldPassRegistration = (
+    !!oldClient &&
+    oldClient.registrationAccessToken !== '' &&
+    newClient.registrationAccessToken === ''
```
aenario: Following https://github.com/cozy/cozy-stack/pull/240 , it should be `== undefined`

jinroh: With [this](https://github.com/cozy/cozy-client-js/blob/master/src/auth_v3.js#L14), we should always have a string in `registrationAccessToken`.

jinroh: I'll update the server mock though to reflect this change and test it.

jinroh:
Rebased to fix the build conflicts

PR #46, opened 2017-01-25
Disk usage

jinroh:
Adds a `cozy.settings` namespace with a `cozy.settings.diskUsage()` method. Doc is updated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,11 @@ before_script:
   - npm explore cozy-data-system coffee commands.coffee test-install client-js ../../test/package.json
   - export NAME=client-js
   - export TOKEN=apptoken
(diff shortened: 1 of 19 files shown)


jinroh:
Tests are failing due to the new permissions system of the stack. We still need to figure out a mechanism to allow testing and development on the stack with permissions activated.

aenario:
We need a cozy-stack CLI function to quickly create a token, may be something like
`TOKEN=$(cozy-stack make-token io.cozy.settings) ` and then provide in integration tests the possiblity to pass the token as a ENV variable.

aenario:
Or (temporarily) use Oauth with proper scope to create said token in tests.

PR #47, opened 2017-01-25
Oauth offline

jinroh:
Makes sure that the `cozy.authorize()` does not fail in case we can not access the client from the server at startup, since should be able to use the stored client and token. The `getClient` is just here to make sure we have an up-to-date client instance but it not necessary to start the application.

This should solve the startup of a mobile application while offline or with bad connectivity at startup.

----------

> If we fall into an error while fetching the client (because of a
> bad connectivity for instance), we do not bail the whole process
> since the client should be able to continue with the persisted
> client and token.
> 
> If it is an explicit Unauthorized error though, we bail, clear th
> cache and retry.


enguerran:
Great idea

src/auth_v3.js
```diff
@@ -237,7 +237,21 @@ export function oauthFlow (cozy, storage, clientParams, onRegistered) {
...
+          // bad connectivity for instance), we do not bail the whole process
+          // since the client should be able to continue with the persisted
+          // client and token.
+          //
+          // If it is an explicit Unauthorized error though, we bail, clear th
+          // cache and retry.
+          if (FetchError.isUnauthorized(err)) {
+            throw err
+          }
+          return { client: oldClient, token }
```
aenario: might be useful to set some variable on the client .isOffline, so the app can know what's going on.

PR #55, opened 2017-02-02
Prevent crash if pouchdb is not included (<script> usage)

aenario:


src/offline.js
```diff
@@ -1,8 +1,9 @@
 import PouchDB from 'pouchdb'
 import pouchdbFind from 'pouchdb-find'
-PouchDB.plugin(pouchdbFind)
 import { DOCTYPE_FILES } from './doctypes'
 
+let pluginLoaded = false
```
enguerran: Why is it a global variable and not a `createDatabase()` variable?

enguerran: Why is `pluginLoaded` a "global" variable and not a `createDatabase()` variable?

src/offline.js
```diff
@@ -11,6 +12,10 @@ export function init (cozy, { options = {}, doctypes = [], timer = false }) {
 }
 
 export function createDatabase (cozy, doctype, options = {}, timer = false) {
+  if (!pluginLoaded) {
```
kosssi: @enguerran pluginLoaded it's used to load only one time pouchdbFind

PR #64, opened 2017-02-10
Add a lastModifiedDate option to directory creation

sebn:
See: https://github.com/cozy/cozy-stack/pull/288

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -324,11 +324,13 @@ It returns a promise for the document of the directory created.
 - `options` is an object with the following fields:
   * `name`: specify the name of the directory
   * `dirID`: specify identifier of the file's directory. if empty, it is the root directory.
+  * `lastModifiedDate`: a date to specify the last modification time to use for the directory.
 
 ```javascript
 const created = await cozy.files.createDirectory({
   name: "mydir",
-  dirID: "123456"
+  dirID: "123456",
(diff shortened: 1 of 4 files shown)


aenario:
Could you make Travis green before we merge this ?

sebn:
Adding some logging to `createDirHandler` in cozy-stack shows that it receives `2017-01-31T20:11:54.147Z`, but sets `doc.CreatedAt` to the current time.

I wonder whether javascript's millisecond precision in `toISOString`'s output could make golang's `time.Parse` fail. Looking at the documentation, the example string for `RFC3339` is `2006-01-02T15:04:05Z07:00`.

We could either:
1. make `cozy-client-js` format dates to match golang's expectation
2. make `cozy-stack` support javascript's `toISOString` format.

The latter may require a bit more work, but it would make life easier for javascript developers writing apps for cozy.

WDYT @aenario @nono ?

sebn:
I noticed the [corresponding cozy-stack unit test](https://github.com/cozy/cozy-stack/blob/master/web/files/files_test.go#L259) uses the same format as javascript Date's `toGMTString()` method.

Replacing `toGMTString()` with `toISOString()` in `cozy-client-js` seems to fix the tests.
Whatever the format we choose, we should not forget to do the same in `cozy.files.create()`.

PR #65, opened 2017-02-11
Fix file creation date

sebn:
Use GMT format since javascript's toISOString doesn't play
well with golang.


sebn:
See https://github.com/cozy/cozy-client-js/pull/64

nono:
-> #68

PR #67, opened 2017-02-13
Make node stream support official in cozy.files.downloadBy*()

sebn:
It already works thanks to node-fetch.

Add some assertions in unit test to make sure body remains
the same upstream, and update documentation accordingly.


sebn:
:ship:

PR #72, opened 2017-02-15
fix: getClient needs only clientID and registrationAccessToken

enguerran:
I don't know why there were an exception thrown if we create a client without `redirectURI`, `softwareID` or `clientName`.
It does not really help when we want to `getClient()`.

But it may be mandatory for `cozy.init()` with oauth, does not it?


nono:
> But it may be mandatory for cozy.init() with oauth, does not it?

It is. I'd have keep the checks, but only when the registration token is missing.

PR #74, opened 2017-02-15
Build UMD modules in root['cozy']

m4dz:
This PR ensure we build a real UMD module, and exposes it at root level under the namespace `cozy` (i.e. `window.cozy.client`)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,5 @@
 /* global fetch */
+
 import {unpromiser, retry, warn} from './utils'
 import {LocalStorage, MemoryStorage} from './auth_storage'
 import {AppToken as AppTokenV2, getAppToken as getAppTokenV2} from './auth_v2'
@@ -91,7 +92,7 @@ const settingsProto = {
   diskUsage: settings.diskUsage
 }
 
-class Cozy {
+class Client {
   constructor (options) {
     this.files = {}
     this.offline = {}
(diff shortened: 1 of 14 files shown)


nono:
@m4dz can you fix the build? In the unit test, there are `cozy = new Cozy` that should be replaced with `cozy = new Client` (and maybe other errors)

src/index.js
```diff
@@ -244,12 +245,6 @@ function addToProto (ctx, obj, proto, disablePromises) {
...
-export default cozy
-export { Cozy, LocalStorage, MemoryStorage }
-
-if ((typeof window) !== 'undefined') {
-  window.cozy = cozy
-  window.Cozy = Cozy
-}
+module.exports = new Client()
+module.exports.LocalStorage = LocalStorage
+module.exports.MemoryStorage = MemoryStorage
```
enguerran: If we use `babel-loader`, why can't we use es2015 `import/export`?

m4dz: We can't, as webpack expect `module.exports` (so, a commonjs format) to build proper UMD. Else, it won't output a `root['cozy']['client'] = Client` but `root['cozy']['client'] = {default: Client}` which isn't what we expect :wink:

enguerran: understood

m4dz:
@nono done, thanks :tada:

nono:
@m4dz does the documentation needs to be updated?

nono:
Conflicts resolved in #75

PR #82, opened 2017-02-21
Fix unregisterClient

nono:


enguerran:
How does this PR integrate @kosssi's #84 ? Or the opposite 😄

PR #83, opened 2017-02-21
[feat] Support more settings API methods

y-lohse:
This PR adds more methods to the settings namespace.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -559,6 +559,25 @@ const usage = await cozy.settings.diskUsage()
 console.log(usage.attributes.used)
 ```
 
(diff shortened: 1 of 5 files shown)


m4dz:
Maybe I'm wrong, but the methods are added to `cozy.client.settings`, not `cozy.settings`. In fact, I wish we can expose those top-level methods under `cozy` namespace rather than `cozy.client` later, but it's not the case yet. Can you update the documentation to reflect it? Thanks!

y-lohse:
You are correct, but actually the whole documentation is now outdated. If you don't mind, can we merge this PR, and then I'll submit a separate PR that updates the whole documentation at once?

m4dz:
Al'right :rocket:

PR #84, opened 2017-02-21
Fix unregister client :bug:

kosssi:


src/auth_v3.js
```diff
@@ -123,7 +123,18 @@ export function updateClient (cozy, client, resetSecret = false) {
 }
 
 export function unregisterClient (cozy, client) {
-  return cozyFetchJSON(cozy, 'DELETE', `/auth/register/${client.clientID}`)
+  if (!(client instanceof Client)) {
+    client = new Client(client)
+  }
+  if (!client.isRegistered()) {
+    return Promise.reject(new Error('Client not registered'))
```
enguerran: I would write *"Client is not registered"*.

src/auth_v3.js
```diff
@@ -123,7 +123,18 @@ export function updateClient (cozy, client, resetSecret = false) {
...
+  }
+  return cozyFetchJSON(cozy, 'DELETE', `/auth/register/${client.clientID}`, null, {
+    manualAuthCredentials: {
+      client: client,
+      token: client
+    }
```
enguerran: You can write:
```javascript
manualAuthCredentials: {
  client,
  token: client
}
```

But what is this object, by the way? How is it necessary to set `client` and `token` to the same value?

nono:
We avoid to build inside the PRs, as it often leads to git conflicts

PR #86, opened 2017-02-22
Fix Travis asynchronous test for offline

enguerran:
[offline.js](./src/offline.js) was [on error](https://travis-ci.org/cozy/cozy-client-js/builds/204656570#L794) with `db.on is not a function` as `db` was the result of a replication and not a replication anymore (asynchronous code).

This error comes from `startSync` inside a `setInterval` function, so asynchronous. So we could not tell mocha to fail if something in this `setInterval` function failed.

I move the `delete replication` instruction on another line of code.

And I rewrite mocha tests to use `should.js` asynchronous and promise API.

Hope it is good for you.

Next step: remove side effect on offline.js code.

### DIFF
--- a/test/integration/offline.js
+++ b/test/integration/offline.js
@@ -20,6 +20,10 @@ let docs = [
   {group: 'B', path: '/a/b/d', year: 2104}
 ]
 
+function sleep (ms) {
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
 describe('offline', function () {
   const cozy = {}
 
@@ -31,17 +35,13 @@ describe('offline', function () {
       cozyURL: COZY_STACK_URL,
       token: COZY_STACK_TOKEN
     })
-    for (var i = 0, l = docs.length; i < l; i++) {
-      docs[i] = await cozy.client.data.create(DOCTYPE, docs[i])
-    }
+    docs = await Promise.all(docs.map(doc => cozy.client.data.create(DOCTYPE, doc)))
   })
 
   after(async function () {
     if (COZY_STACK_VERSION === '3') {
-      for (var i = 0, l = docs.length; i < l; i++) {
-        await cozy.client.data.delete(DOCTYPE, docs[i])
-      }
-      cozy.client.offline.stopAllSync()
+      await docs.forEach(doc => cozy.client.data.delete(DOCTYPE, doc))
+      cozy.client.offline.stopAllSync('after')
     }
   })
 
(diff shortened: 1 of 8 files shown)


test/integration/offline.js
```diff
@@ -53,66 +53,55 @@ describe('offline', function () {
...
-  }).timeout(7 * 1000)
+  it('can sync and stop sync a database', async function () {
+    // create a database
+    const db = cozy.client.offline.createDatabase(DOCTYPE, { adapter: 'memory' })
+    // create a doc
+    const sampleDoc = { data: 'some Data' }
+    const remoteDoc = await cozy.client.data.create(DOCTYPE, sampleDoc)
+    // check the db to look for the new doc
+    db.get(remoteDoc._id).should.be.rejectedWith({ message: 'missing' })
+    // activate synchronisation every 1000ms
```
kosssi: 100ms ;)

test/integration/offline.js
```diff
@@ -31,17 +35,13 @@ describe('offline', function () {
...
   })
 
   after(async function () {
     if (COZY_STACK_VERSION === '3') {
-      for (var i = 0, l = docs.length; i < l; i++) {
-        await cozy.client.data.delete(DOCTYPE, docs[i])
-      }
-      cozy.client.offline.stopAllSync()
+      await docs.forEach(doc => cozy.client.data.delete(DOCTYPE, doc))
+      cozy.client.offline.stopAllSync('after')
```
kosssi: Can you remove 'after' ;)

y-lohse:
Sorry, I don't think I can do a good review of these changes, there are too many things I'm not familiar with. I do understand it solves the promise-related bug we discussed, and I see how it fixes it, but beyond that... ¯\_(ツ)_/¯

test/integration/offline.js
enguerran: Oh yes, it was for debug async calls 😜

enguerran:
I cannot finish the work initiated here as I cannot understand how does the pouchDB replication work.

I success to run integration tests on my local environment (see #93 ) and I try to refactor the code to fix the unhandled `db.on is not a function` error on travis' builds.

screenshot at 13-24-09
 
First I tried to let travis fail with the actual code base.

Then I tried to make `startSync`'s `setInterval`'s promise to be managed outside the `offline.js` API that could let `mocha-webpack` to wait for those instructions to finish before it ends itself.

Finally I could not understand how [pouchDB replication](https://pouchdb.com/api.html#replication) works as I have to wait a while (using the sleep function) for the replication to be complete. And I saw that two **consecutive** tests result in different information in the replication's result object:

screenshot at 13-30-39

Then

screenshot at 13-30-58

300 or 722 docs read and written but when I query the cozy-stack's couchDB, I can see only 2 docs.

```
http http://cozy.local:8080/data/io.cozy.testobject2/_all_docs\?include_docs\=true 
```

I admit my sad defeat and let chances to others. 😞

enguerran:
@kosssi are we ok to close this PR as #94 replaced it?

kosssi:
it's merged in https://github.com/cozy/cozy-client-js/pull/94 ;) thanks

PR #90, opened 2017-02-23
Fix refresh token

nono:


y-lohse:
Purely looking at the code, it's a 👍🏼 for me, but I don't know the context of these changes. Do they need to be double-checked? If so, would you mind describing a testing scenario?

nono:
> Do they need to be double-checked?

No, it's okay. Thanks for the review

PR #93, opened 2017-02-27
readme: how to run integration tests on local environment

enguerran:
I had hard time to run integration tests on my local environment.
But as it is very useful, you may find the workflow I use.

If it can interest and help anyone, be my guess.


nono:
I think it is really helpful. Maybe, we can put that in CONTRIBUTING.md instead of README.md

PR #96, opened 2017-02-28
Fix revoked client

nono:


src/fetch.js
```diff
@@ -99,19 +99,25 @@ function handleJSONResponse (res) {
   }
 }
 
-export class FetchError {
+export class FetchError extends Error {
   constructor (res, reason) {
+    super()
+    Error.captureStackTrace(this, this.constructor)
```
jinroh: I think `Error.captureStackTrace` is not standard (not available in firefox). We should have a guard `if (Error.captureStackTrace) { .. }`

src/auth_v3.js
```diff
@@ -281,8 +281,8 @@ export function oauthFlow (cozy, storage, clientParams, onRegistered) {
           //
           // If it is an explicit Unauthorized error though, we bail, clear th
           // cache and retry.
-          if (FetchError.isUnauthorized(err)) {
-            throw err
+          if (FetchError.isUnauthorized(err) || FetchError.isNotFound(err)) {
+            throw new Error('Client has been revoked')
```
enguerran: Why don't you throw a `new FetchError`?

nono: I don't consider that a revoked client is a `FetchError`: it is not really tied to a single request to the stack, but something more global.

enguerran: Sounds good to me.

PR #99, opened 2017-03-03
[feat] Add helper to get file path

gregorylegarec:
To call `cozy.files.getDownloadLink`, I needed to generate path for a given file.

As this little piece of logic is already used in Files, I think we should add a little helper in cozy-client-js to facilitates further developments and mutualize code.


aenario:
:+1: We might add a File model in the future wich will extends jsonapi.Object to carry this kind of method

PR #105, opened 2017-03-06
handle revocation in getClient

enguerran:
The revocation is only handle in the oauthFlow, I think it is to late to catch the error.
Maybe we can remove the fix by @nono on line 287.

### DIFF
--- a/src/auth_v3.js
+++ b/src/auth_v3.js
@@ -149,12 +149,26 @@ export function getClient (cozy, clientParams) {
   if (!cli.isRegistered()) {
     return Promise.reject(new Error('Client not registered'))
   }
-  return cozyFetchJSON(cozy, 'GET', `/auth/register/${cli.clientID}`, null, {
-    manualAuthCredentials: {
-      token: cli
-    }
-  })
+  return cozyFetchJSON(cozy, 'GET', `/auth/register/${cli.clientID}`, null,
+    {
+      manualAuthCredentials: {
+        token: cli
+      }
+    })
     .then((data) => createClient(data, cli))
+    .catch((err) => {
+      // If we fall into an error while fetching the client (because of a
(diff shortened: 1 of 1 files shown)


enguerran:
It is mandatory for https://github.com/cozy/cozy-files-v3/pull/138

src/auth_v3.js
```diff
@@ -149,12 +149,25 @@ export function getClient (cozy, clientParams) {
...
+    })
+    .catch((err) => {
+      // If we fall into an error while fetching the client (because of a
+      // bad connectivity for instance), we do not bail the whole process
+      // since the client should be able to continue with the persisted
+      // client and token.
+      //
+      // If it is an explicit Unauthorized error though, we bail, clear th
+      // cache and retry.
+      if (FetchError.isUnauthorized(err) || FetchError.isNotFound(err)) {
```
nono: `cozyFetchJSON` is used for many things and a 404 error does not mean that the client has been revoked in most cases

enguerran: Ok, i'll find another way.

enguerran: @nono:
I put the `catch` code in `getClient` https://github.com/cozy/cozy-client-js/blob/master/src/auth_v3.js#L147-L158 not in `cozyFetchJSON`.
Do you really think there is many things and a 404 error in `getClient` could be something else than a revocation?

nono: If `getClient` returns a 404, it means that the client does not exist. Either it has been revoked, or the client has never existed on this instance (if you take the client_id of one instance and call `getClient` on another instance for example). The effect is similar: you can't use the client and you should treat that as a revoked client.

enguerran:
Just realized I erase the line https://github.com/cozy/cozy-client-js/blob/master/src/auth_v3.js#L157...

Now I fixed it and the build is happier.

😜

PR #109, opened 2017-03-08
feat: enhance offline behavior

enguerran:
onError and onComplete are both being called too early.

And I don't believe it is a good idea to raise a *no database error* if we start a replication.
I do believe that when we launch a replication, if there is no local database, we should create one.

It is related to https://github.com/cozy/cozy-files-v3/pull/157


enguerran:
Ohh, well, I need to fix some test as they're out of scope now. Stay tuned!

PR #111, opened 2017-03-09
Get secret link from two ways: path or Id

CPatchane:


src/index.js
```diff
@@ -66,7 +66,8 @@ const filesProto = {
   statByPath: files.statByPath,
   downloadById: files.downloadById,
   downloadByPath: files.downloadByPath,
-  getDownloadLink: files.getDownloadLink,
```
gregorylegarec: What do you think about keeping `getDownloadLink` pointing to `getDownloadLinkByPath`, to avoid breaking a lot of things in Files and Photos ? We could comment it as `deprecated` and remove it in a couple of release.

CPatchane: @gregorylegarec You're right, I can quickly fix in Photos but maybe for Files that can be more complicated in this sprint's end. I'll add it 👍

gregorylegarec:
Thanks @CPatchane 👍

PR #118, opened 2017-03-23
test: update tests with built-in functions

enguerran:
I take this opportunity to ask if this very test is really interesting.
I think it test pouchDB and not very much the library's code.

If we want to be sure the replication works, we only have to check
if the `offline` document exists (`!== undefined`), don't we?


kosssi:
I prefer that we keep the test like this. It allows to check the differences with the rendering of the stack. What do you think @aenario ?

PR #119, opened 2017-03-23
fix: mime-type was not detected in cordova

enguerran:
When a cordova application uses the cordova-plugin-file plugin, `window.File` is overwritten by another object, see https://github.com/apache/cordova-plugin-file/blob/master/www/File.js#L31.

So the line `data instanceof File` is `false` then the code enters the `isBlob` conditionnal scope that is why we change this line to get `data.type` if it exists or `contentTypeOctetStream` otherwise.

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -35,7 +35,7 @@ function doUpload (cozy, data, method, path, options) {
         lastModifiedDate = data.lastModifiedDate
       }
     } else if (isBlob) {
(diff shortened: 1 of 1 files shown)


enguerran:
Don't be afraid, @reviewers, I don't know who is the best one to review this PR. If you think you are not qualified, just remove your name.

Thanks for your attention.

PR #123, opened 2017-03-31
[fix] Make sure FetchError has a proper message

### LINKED ISSUE #112: Error on invalid query should be more explicit
## Summary

Calling cozy.client.data.query with invalid parameters should return an explicit error message

## STR
```
cozy.client.data.query(myIndex, {}).then().catch((err) => { console.error(err);})
```

## Expected result

The full error returned by the server: `{"error":"missing_required_key","ok":false,"reason":"Missing required key: selector","status":"400"}`

## Actual result

Only an Error object, without name nor message.

## Version of CozyClientJs

0.1.7

sebn:
Should fix #112

### DIFF
--- a/src/fetch.js
+++ b/src/fetch.js
@@ -111,6 +111,11 @@ export class FetchError extends Error {
     this.url = res.url
     this.status = res.status
     this.reason = reason
+
+    Object.defineProperty(this, 'message', {
+      value: reason.message || `${this.name}: ` +
+        (typeof reason === 'string' ? reason : JSON.stringify(reason))
+    })
   }
 }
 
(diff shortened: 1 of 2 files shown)


sebn:
Appvengers, any thought? 😃

sebn:
Oops, forgot you were one of them @y-lohse.
Merging then.

sebn:
@y-lohse I believe this issue appeared when we started subclassing `Error`. Extending built-in classes is officially unsupported in Babel and generally doesn't work as intended, requiring various fixes.

y-lohse:
Gotcha, thanks!

PR #124, opened 2017-03-31
[fix] Don't duplicate FetchError name in message

sebn:


sebn:
Sorry, I messed up with the message (was something like `FetchError: FetchError: blah`). Fixed now.

PR #125, opened 2017-03-31
Use cozy.tools instead of cozy.local (no need to edit /etc/hosts)

nono:
Cf https://github.com/cozy/cozy-stack/pull/435


gregorylegarec:
Thanks @nono 👍

PR #126, opened 2017-04-03
feat: Add intents

gregorylegarec:
This PR adds `cozy.client.intents.start()` methods wich allow to inject an iframe from another service for a given intent.

`cozy.client.intents.resolve()`, used in services, will be implemented later.

At this time, it is almost a POC. As we don't have a service to get the proper url from a given intent, the `cozy.client.intents.start()` method takes a temporary third argument, which is the mocked url that we want to use for the intent.

It has been discussed with @y-lohse where we should implement the intents logic.

First, it as been proposed that they may be served by the cozy-bar. But as intents are needed both in onboarding and in intent service pages, which do not contain a cozy-bar, we decided to implement them in cozy-client-js. Our fear was that DOM interaction would have been too much for a lib dedicated to call the stack. Eventually, it appeared that the main interaction was injecting an iframe and use the `window.postMessage` API. As this is a small bunch of code, I feel that it fit well in `cozy-client-js`.

### DIFF
--- a/src/intents.js
+++ b/src/intents.js
@@ -1,3 +1,4 @@
+/* global URL */
 const intentClass = 'coz-intent'
 
 function getUrl (intent, temporaryMockedIntentUrl) {
@@ -17,7 +18,7 @@ function injectService (url, element, intent) {
   element.appendChild(iframe)
 
   // Keeps only http://domain:port/
-  const serviceOrigin = url.split('/', 3).join('/')
+  const serviceOrigin = new URL(url).origin
 
   return new Promise((resolve, reject) => {
     let handshaken = false
--- a/src/intents.js
+++ b/src/intents.js
@@ -1,0 +1,61 @@
+const intentClass = 'coz-intent'
+
+function getUrl (intent, temporaryMockedIntentUrl) {
+  return Promise.resolve(temporaryMockedIntentUrl)
+}
+
+function injectService (url, element, intent) {
+  const document = element.ownerDocument
+  if (!document) throw new Error('Cannot retrieve document object from given element')
+
+  const window = document.defaultView
+  if (!window) throw new Error('Cannot retrieve window object from document')
+
+  const iframe = document.createElement('iframe')
+  iframe.setAttribute('src', url)
+  iframe.classList.add(intentClass)
+  element.appendChild(iframe)
+
+  // Keeps only http://domain:port/
+  const serviceOrigin = url.split('/', 3).join('/')
+
+  return new Promise((resolve, reject) => {
+    let handshaken = false
+    const messageHandler = (event) => {
+      if (event.origin !== serviceOrigin) return
+
+      if (event.data === 'intent:ready') {
+        handshaken = true
+        return event.source.postMessage(intent.data, event.origin)
+      }
+
+      window.removeEventListener('message', messageHandler)
+
+      if (event.data === 'intent:error') {
+        return reject(new Error('Intent error'))
+      }
+
+      return handshaken
+        ? iframe.parentNode.removeChild(iframe) && resolve(event.data)
+        : reject(new Error('Unexpected handshake message from intent service'))
+    }
+
+    window.addEventListener('message', messageHandler)
+  })
+}
+
+export function start (client, intent, element, temporaryMockedIntentUrl) {
+  ['action', 'type'].forEach((property) => {
+    if (!intent[property]) {
+      throw new Error(`Misformed intent, "${property}" property must be provided`)
+    }
+  })
+
+  // To replace with a call to client
+  return getUrl(intent, temporaryMockedIntentUrl)
+    .then(url => injectService(url, element, intent))
+}
+
+export function resolve () {
+  // TODO
+}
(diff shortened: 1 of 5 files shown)


src/intents.js
```diff
@@ -0,0 +1,61 @@
...
+function injectService (url, element, intent) {
+  const document = element.ownerDocument
+  if (!document) throw new Error('Cannot retrieve document object from given element')
+
+  const window = document.defaultView
+  if (!window) throw new Error('Cannot retrieve window object from document')
+
+  const iframe = document.createElement('iframe')
+  iframe.setAttribute('src', url)
+  iframe.classList.add(intentClass)
```
y-lohse: I can't see where `ìntentClass` comes from?

src/intents.js
```diff
@@ -0,0 +1,61 @@
...
+        ? iframe.parentNode.removeChild(iframe) && resolve(event.data)
+        : reject(new Error('Unexpected handshake message from intent service'))
+    }
+
+    window.addEventListener('message', messageHandler)
+  })
+}
+
+export function start (client, intent, element, temporaryMockedIntentUrl) {
+  ['action', 'type'].forEach((property) => {
```
y-lohse: Just wondering, have you considered making `start`'s signature more verbose?

`intent` in this case is going to be a hashtable with up to 4 properties ([`action`, `type`, `data` and `permissions`](https://github.com/cozy/cozy-stack/blob/master/docs/intents.md#2-intent-start)). Personnaly, I don't like when an API has to handle a "black box" like that. For example, `intent.data` is used later on, but it's impossible to tell just by looking at the function definition. Naïvely, I would go with something like `start(client, element, action, type, data = {}, permissions = [], temporaryMockedIntentUrl)`.

src/intents.js
```diff
@@ -0,0 +1,61 @@
...
+        return reject(new Error('Intent error'))
+      }
+
+      return handshaken
+        ? iframe.parentNode.removeChild(iframe) && resolve(event.data)
+        : reject(new Error('Unexpected handshake message from intent service'))
+    }
+
+    window.addEventListener('message', messageHandler)
+  })
```
y-lohse: Neat. I especially like that you were able to return a promise here 👍

src/intents.js
```diff
@@ -0,0 +1,61 @@
...
+  const window = document.defaultView
+  if (!window) throw new Error('Cannot retrieve window object from document')
+
+  const iframe = document.createElement('iframe')
+  iframe.setAttribute('src', url)
+  iframe.classList.add(intentClass)
+  element.appendChild(iframe)
+
+  // Keeps only http://domain:port/
+  const serviceOrigin = url.split('/', 3).join('/')
```
nono: Maybe

```js
const serviceOrigin = new URL(url).origin
```

src/intents.js
gregorylegarec: Defined at the top of the file. ⬆️

src/intents.js
gregorylegarec: I see your point. Personally I prefer passing an object because I think it is more robust and extensible, but less clear I admit. I'll try to do something. After a couple of other PR. :)

y-lohse: Don't feel forced to :-). It's a personal thing, as long as it's a conscious decision I'm good with it!

gregorylegarec: @y-lohse I thank about it. Actually what bothers me is to have `element` as first parameter, I think `intent`, or intent's properties have to be in first position. But then, I think that we got to much params, with four related to the intent, and one last (`element`), absolutely not. So maybe we could make something like:
```javascript
cozy.client.intents.create('PICK', 'io.cozy.files).start(element).then(document => {})
//or
cozy.client.intents.create('EDIT', 'io.cozy.contact).then(intent => cozy.client.intents.start(intent, element).then(document => {})
```
By the way, a time will come when we will need to create an intent in DB. Maybe this syntax will be the best one.

y-lohse: I don't mind the first one. In the second one, the intent creation is async which there is no reason for.

> By the way, a time will come when we will need to create an intent in DB.

Really? Then yes, the first one makes even more sense.

gregorylegarec: Yep, see
https://github.com/cozy/cozy-stack/blob/master/docs/intents.md#post-intents
and
https://github.com/cozy/cozy-stack/blob/master/docs/intents.md#get-intentsid

Maybe we should already create it in this PR.

y-lohse: Oh, *that part*. Well, creating the intent in the DB and getting the service URL is really one and the same step, so I don't think there will be more than one function. But yeah, our first proposal makes more and more sense.
Do you want to change the current PR to include it, or would you rather sit on it for a while and maybe do it later? I'm totally fine with both.

gregorylegarec: I can totally live with the idea to do it later, when we will have clearer view. 👍

gregorylegarec: Mmmmh actually, with last developements in, I need to save the intent to be able to display the iframe. Let me update my PR after lunch break. :)

PR #128, opened 2017-04-06
[WIP] Need docs : add intents service

gregorylegarec:


src/intents.js
```diff
@@ -18,7 +16,7 @@ function injectService (url, element, intent) {
   element.appendChild(iframe)
 
   // Keeps only http://domain:port/
-  const serviceOrigin = new URL(url).origin
+  const serviceOrigin = url.split('/', 3).join('/')
```
y-lohse: Is that revert on purpose?

src/intents.js
```diff
@@ -27,36 +25,86 @@ function injectService (url, element, intent) {
...
 }
 
-export function resolve () {
-  // TODO
+// returns a service to communicate with intent client
+export function createService (cozy, id, window) {
+  return cozyFetchJSON(cozy, 'GET', `/intents/${id}`)
+    .then(intent => {
+      return listenClientData(intent, window)
+        .then(data => {
```
y-lohse: This whole thing is really nice, but is there any particular reason why you don't chain the promises instead of nesting them? ie.

```
export function createService (cozy, id, window) {
  return cozyFetchJSON(cozy, 'GET', `/intents/${id}`)
    .then(intent => {
      return listenClientData(intent, window)
    })
    .then(data => {
          let terminated = false
          return etc...
     })
}
```

gregorylegarec: Yep, at first I chained them, but then I realised that I needed the `intent` at least in the `terminate()` method. So I chained this way to keep the scope.

src/intents.js
gregorylegarec: `URL` does not exist in node, so tests were failing. After a couple of minutes looking for a polyfill, I gave up.

src/intents.js
y-lohse: Aaah yes, makes sense.

gregorylegarec:
@y-lohse Je le note ici : je pense qu'il sera intéressant de gérer un timeout sur le handshake.

Issue #88, opened 2017-02-23
Pouchdb as a dependency


y-lohse:
We are using pouch in the [offline module](https://github.com/cozy/cozy-client-js/blob/master/src/offline.js#L1), but it's declared as a [dev dep](https://github.com/cozy/cozy-client-js/blob/master/package.json#L53), so it's not always installed. I believe it should be declared as a regular dependency.

I'd be happy to supply a PR but I'm not sure how we should declare `pouchdb-find` and `pouchdb-adapter-memory`, as normal deps or as peer deps?

Ping @kosssi and @enguerran

kosssi:
I'm ok to load pouchdb on dep of cozy-client-js.

In the last release of [pouchdb](https://github.com/pouchdb/pouchdb/releases/tag/6.1.2) pouchdb-find is include so is not necessary ;)

 pouchdb-adapter-memory is specific to test cozy-client-js offline.

y-lohse:
Ok, thanks! Following our conversation, just pinging @m4dz to be sure: pouchdb isn't very big, but this change would still mean we'd bundle it along with cozy-client-js, as far as I'm aware. Is that ok?

I mean, if it's used in the module, I strongly believe it should be declared as a regular dependency, but let's get everyone's input on this.

nono:
It is 38MB to download. I'd prefer to keep it as an optional dependency.

y-lohse:
So there are 2 aspects here: 

1. if you use the offline part of cozy-client-js, you *need* pouchdb installed somewhere. IMO this shouldn't be the host package's job, but why not.
2. If an app imports cozy-client-js as part of it's build process but doesn't have pouchdb as a dependency, the build breaks. This is happening with photos at the moment. Adding pouchdb as a dependency to photos, despite not using it in any way, seems wrong. But maybe there's a way to fix the build without pouchdb?

enguerran:
Are [peerDependencies](https://nodejs.org/en/blog/npm/peer-dependencies/#the-solution-peer-dependencies) not a good solution for such a library?

What is the wanted behavior?

- you don't want to use any kind of local persistence ⇒ no pouchDB is added, and the build do not break
- you want to use local persistence ⇒ you need to add pouchDB somewhere in your host bundle

The fact `cozy-photo-v3`'s build breaks if there is no pouchDB installed is not good and we may fix `cozy-client-js` to not require pouchDB mostly if we don't want local persistence.

y-lohse:
[...]

PR #131, opened 2017-04-11
chore: removes PouchDB dependency

gregorylegarec:
Should fix the issue discussed in https://github.com/cozy/cozy-client-js/issues/88.

Instead of editing [`webpack.target.mobile.js`](https://github.com/cozy/cozy-files-v3/blob/master/config/webpack.target.mobile.js) like @m4dz proposal, I suggest to use `webpack.ProvidePlugin()` instead (Actually using  `imports-loader` did not work for me) :

```js
module.exports = {
  ...
  plugins: [
    new webpack.ProvidePlugin({
       'PouchDB': 'pouchdb',
       'pouchdbFind': 'pouchdb-find'
    })
  ]
}
```

### DIFF
--- a/docs/offline.md
+++ b/docs/offline.md
@@ -1,0 +1,27 @@
+This document describes in more details how to use this library in offline mode.
+
+All applications do not require offline mode, and offline mode needs two large dependencies : `pouchdb` and `pouchdb-find`.
+As those two dependencies are expensive in size, we decided to not provide them by default, since v0.2.0.
+
+So if an app needs to use offline mode, it will need to add those dependencies by itself. It means that developers just need to do:
+
+```bash
+yarn add pouchdb
+yarn add pouchdb-find
+```
+
+Then they will have to import `pouchdb` and `pouchdbfind` globally, as they are not imported in codebase anymore. This means adding the following configuration in all concerned webpack target files (for example `webpack.target.mobile.js`) :
+
+```js
+module.exports = {
+  ...
+  plugins: [
+    new webpack.ProvidePlugin({
+      'PouchDB': 'pouchdb',
+      'pouchdbFind': 'pouchdb-find'
+    })
+  ]
+}
+```
+
+This same configuration has to be added in file `wepback.config.prod.js`, to make the build possible.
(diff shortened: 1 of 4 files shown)


gregorylegarec:
@enguerran Thanks for your feedback, I think I will add a little documentation before merging.

y-lohse:
The dependencies are still listed as `devDependencies` and not `peerDependencies`, is that on purpose? I'm still unclear about the difference.

> I think I will add a little documentation before merging.

Yes, please add a big fat warning in the changelog and maybe how to upgrade. I believe this change is going to break apps that use the `offline` module so we need to make sure they see it coming :-)

(Thanks for taking care of this <3)

gregorylegarec:
@y-lohse PouchDB is at least needed in `devDependencies` for tests, I think.

gregorylegarec:
@y-lohse To highlight the way we are removing a dependency, I suggest to update the next version of cozy-client-js from 0.1.11 to 0.2.0. What do you think ?

y-lohse:
Yes, I agree; That's as close to semver as we'll get right now.

docs/offline.md
```diff
@@ -0,0 +1,27 @@
+This document describes in more details how to use this library in offline mode.
+
+All applications do not require offline mode, and offline mode needs two large dependencies : `pouchdb` and `pouchdb-find`.
+As those two dependencies are expensive in size, we decided to not provide them by default, since v0.2.0.
```
enguerran: s/As those two dependencies are expensive in size/As both are very large/

enguerran: s/we decided to not provide/we decide not to provide/

docs/offline.md
```diff
@@ -0,0 +1,27 @@
...
+
+So if an app needs to use offline mode, it will need to add those dependencies by itself. It means that developers just need to do:
```
enguerran: s/if an app needs/if your app needs/
s/it will need to/you have to/
s/by itself/by yourself/
s/It means that developers just need/It means you just need/

enguerran: When you say `yarn add pouchdb`, do you mean in the app directory or in the lib directory?

gregorylegarec: App directory. I'll precise that.

gregorylegarec:
Fix #88.

PR #133, opened 2017-04-14
feat: prevent intent URL encoding

gregorylegarec:
In production stack seems to encode URLs


gregorylegarec:
Should be fixed by https://github.com/cozy/cozy-stack/pull/487

PR #134, opened 2017-04-17
feat: Add cozyVersion parameter to remove '/status/' request (offline) :bath:

kosssi:
This feature add cozyVersion parameter.

The purpose is to delete [the request to the server '/status/'](https://github.com/cozy/cozy-client-js/blob/88a63b5b052fbe78e9e1835dc4dd61d1ab8c6504/src/index.js#L238) for the offline mode.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -141,7 +141,7 @@ class Client {
     this._authstate = AuthNone
     this._authcreds = null
     this._storage = null
-    this._version = null
+    this._version = options.version || null
     this._offline = null
 
     const token = options.token
@@ -235,7 +235,7 @@ class Client {
 
   isV2 () {
     if (!this._version) {
-      this._version = retry(() => fetch(`${this._url}/status/`), 3)()
+      return retry(() => fetch(`${this._url}/status/`), 3)()
         .then((res) => {
           if (!res.ok) {
             throw new Error('Could not fetch cozy status')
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -141,7 +141,7 @@ class Client {
     this._authstate = AuthNone
     this._authcreds = null
     this._storage = null
-    this._version = null
+    this._cozyVersion = options.cozyVersion || null
```
enguerran: Why do you change the variable name? I think `version` is good enough.

src/index.js
```diff
@@ -234,18 +234,21 @@ class Client {
...
           }
         })
-        .then((status) => status.datasystem !== undefined)
+        .then((status) => {
+          this._cozyVersion = status.datasystem !== undefined ? 2 : 3
+          return this.isV2()
+        })
     }
-    return this._version
+    return Promise.resolve(this._cozyVersion === 2)
```
enguerran: `return` statements seem weird:

- you return a `Promise.resolve` or
- a `retry()` promise

Before those changes the code returns a value. It is not the same type.

enguerran:
Needed for https://github.com/cozy/cozy-drive/pull/234

src/index.js
kosssi: ok

src/index.js
kosssi: all function call is `cozy.isV2().then((isV2) => {` so it's better 😊

enguerran: if all functions wait for a `Promise`, then it will return a `Promise` :+1:

PR #135, opened 2017-04-17
Checks if there is a connection before starting replication :ok_hand:

kosssi:
I'm using [navigator.onLine](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine) to check the online status of the browser.

If no connection we don't launch replication.

I'm test on Android 7 and iOS 10.2.

### DIFF
--- a/src/offline.js
+++ b/src/offline.js
@@ -141,7 +141,8 @@ export function replicateFromCozy (cozy, doctype, options = {}) {
 
     if (isOffline()) {
       reject(replicationOfflineError)
-      return options.onError && options.onError(replicationOfflineError)
+      options.onError && options.onError(replicationOfflineError)
+      return
     }
 
     getReplicationUrl(cozy, doctype)
@@ -218,7 +219,8 @@ export function startRepeatedReplication (cozy, doctype, timer, options = {}) {
   return setRepeatedReplication(cozy, doctype, setInterval(() => {
     if (isOffline()) {
       // network is offline, replication cannot be launched
-      return console.info(replicationOfflineError)
+      console.info(replicationOfflineError)
+      return
     }
     if (!hasReplication(cozy, doctype)) {
       replicateFromCozy(cozy, doctype, options)
--- a/src/offline.js
+++ b/src/offline.js
@@ -1,6 +1,9 @@
 /* global PouchDB, pouchdbFind */
 import {DOCTYPE_FILES} from './doctypes'
 import {refreshToken} from './auth_v3'
+import {isOffline} from './utils'
+
+export const replicationOfflineError = 'Replication abort, your device is actually offline.'
 
 let pluginLoaded = false
 
@@ -136,6 +139,11 @@ export function replicateFromCozy (cozy, doctype, options = {}) {
       return reject(new Error('You can\'t use `live` option with Cozy couchdb.'))
     }
 
+    if (isOffline()) {
+      reject(replicationOfflineError)
+      return options.onError && options.onError(replicationOfflineError)
+    }
+
     getReplicationUrl(cozy, doctype)
       .then(url => setReplication(cozy, doctype,
         getDatabase(cozy, doctype).replicate.from(url, options).on('complete', (info) => {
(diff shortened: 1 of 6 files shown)


src/offline.js
```diff
@@ -136,6 +139,11 @@ export function replicateFromCozy (cozy, doctype, options = {}) {
       return reject(new Error('You can\'t use `live` option with Cozy couchdb.'))
     }
 
+    if (isOffline()) {
+      reject(replicationOfflineError)
+      return options.onError && options.onError(replicationOfflineError)
```
enguerran: In a promise function, calling `resolve` or `reject` won't let subsequent code to be executed. In other words, the instruction `return options.onError && options.onError(replicationOfflineError)` won't be executed, never.

src/offline.js
```diff
@@ -208,6 +216,10 @@ export function startRepeatedReplication (cozy, doctype, timer, options = {}) {
   }
 
   return setRepeatedReplication(cozy, doctype, setInterval(() => {
+    if (isOffline()) {
+      // network is offline, replication cannot be launched
+      return console.info(replicationOfflineError)
```
enguerran: `console.info()` don't return anything: https://developer.mozilla.org/en-US/docs/Web/API/Console/info

What was your intention?

src/utils.js
```diff
@@ -23,6 +24,22 @@ export function isPromise (value) {
...
 
+export function isOnline () {
+  if (navigator === undefined) {
+    return undefined
+  }
+
+  return navigator.onLine === true
+}
+
+export function isOffline () {
```
enguerran: Regarding the code duplication and the needs, don't you think it is too much code.

```js
export function isOnline () {
  return navigator ? navigator.onLine : undefined
}

export function isOffline () {
  return !isOnline()
}
```

Or with even less code, you define only one function and the developers could use `isOnline()` or `!isOnline()` instead of `isOnline()` and `isOffline()`.

test/integration/files.js
```diff
@@ -12,6 +12,8 @@ const COZY_STACK_URL = process.env && process.env.COZY_STACK_URL || ''
 const COZY_STACK_VERSION = process.env && process.env.COZY_STACK_VERSION
 const COZY_STACK_TOKEN = process.env && process.env.COZY_STACK_TOKEN
 
+global.navigator = undefined
```
enguerran: Why do you need to instantiate `gobal.navigator` to `undefined`.
If you don't declare `global.navigator`, then `global.navigator` is `undefined`...

test/integration/offline.js
```diff
@@ -13,6 +13,7 @@ PouchDB.plugin(require('pouchdb-adapter-memory'))
 // environment, so we declare it in global scope here.
 global.PouchDB = PouchDB
 global.pouchdbFind = pouchdbFind
+global.navigator = undefined
```
enguerran: Why do you need to instantiate `gobal.navigator` to `undefined`.
If you don't declare `global.navigator`, then `global.navigator` is `undefined`...

test/unit/auth.js
```diff
@@ -7,6 +7,8 @@ import {oauthFlow, AccessToken} from '../../src/auth_v3'
 import mock from '../mock-api'
 import {decodeQuery} from '../../src/utils'
 
+global.navigator = undefined
```
enguerran: Why do you need to instantiate `gobal.navigator` to `undefined`.
If you don't declare `global.navigator`, then `global.navigator` is `undefined`...

enguerran: see this as an example of my beliefs: https://jsbin.com/voxebe/edit?js,console

test/integration/offline.js
enguerran: see this as an example of my beliefs: https://jsbin.com/voxebe/edit?js,console

test/integration/files.js
enguerran: see this as an example of my beliefs: https://jsbin.com/voxebe/edit?js,console

kosssi: I don't know why but test failed if I remove this line ;) https://travis-ci.org/cozy/cozy-client-js/builds/222804829#L640

test/integration/offline.js
kosssi: I don't know why but test failed if I remove this line ;) https://travis-ci.org/cozy/cozy-client-js/builds/222804829#L640

test/unit/auth.js
kosssi: I don't know why but test failed if I remove this line ;) https://travis-ci.org/cozy/cozy-client-js/builds/222804829#L640

src/offline.js
enguerran: I was wrong, it seems to be executed, but the return value is not what you think it would be: https://jsbin.com/xuyayot/1/edit?js,console

PR #136, opened 2017-04-19
Break an infinite loop of retry/failure 

nono:
When fetch returns a 400 but client is not revoked, the client was starting an infinite loop of tries/failures

### DIFF
--- a/src/fetch.js
+++ b/src/fetch.js
@@ -32,7 +32,7 @@ function cozyFetchWithAuth (cozy, fullpath, options, credentials) {
     cozy.isV2(),
     fetch(fullpath, options)
   ]).then(([isV2, res]) => {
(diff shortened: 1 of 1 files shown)


src/fetch.js
```diff
@@ -41,6 +41,7 @@ function cozyFetchWithAuth (cozy, fullpath, options, credentials) {
     if (!client || !(token instanceof AccessToken)) {
       return res
     }
+    options.dont_retry = true
```
gregorylegarec: Looks ok to me, I just think that the property should be camelCase, i.e. `dontRetry` instead of `dont_retry`.

nono: Right, I fix it

kosssi:
I integrate this fix on the next release

PR #138, opened 2017-04-21
[WIP] feat: add cozy.client.accounts.create

gregorylegarec:
No need for tests as we are only encapsulating a `POST` request to `/data/io.cozy.accounts`, which is already tested.

### DIFF
--- a/src/accounts.js
+++ b/src/accounts.js
@@ -1,13 +1,13 @@
 import {cozyFetchJSON} from './fetch'
 
-export function create (cozy, konnector, login = '', password = '', name = '') {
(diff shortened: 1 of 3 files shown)


src/accounts.js
```diff
@@ -0,0 +1,12 @@
+import {cozyFetchJSON} from './fetch'
+
+export function create (cozy, type, login = '', password = '') {
```
nono: Instead of login and password, can we pass an `auth` object in last parameter?

gregorylegarec:
This code will eventually stay in cozy-my-accounts for now.
Maybe we should discuss if `cozy-client-js` should offer this kind of helpers.

PR #142, opened 2017-04-26
feat: adds cozy.client.files.createDirectoryRecursively()

gregorylegarec:
Very useful for creating konnectors folder.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -352,6 +352,19 @@ const created = await cozy.client.files.createDirectory({
 ```
 
 
+### `cozy.client.files.createDirectoryRecursively(path)`
+
+`cozy.client.files.createDirectory(path)` is used to create one or more directories for a given path.
+
+It returns a promise for the document of the last directory created.
+
+- `path` is a string
+
+```javascript
+const barDirectory = await cozy.client.files.createDirectoryRecursively('/Foo/Bar')
+```
+
+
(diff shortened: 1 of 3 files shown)


gregorylegarec:
@Cpatchane I totally agree but I lack the time to write tests. Maybe I should add a TODO aboit that ?

docs/README.md
```diff
@@ -352,6 +352,19 @@ const created = await cozy.client.files.createDirectory({
 ```
 
 
+### `cozy.client.files.createDirectoryRecursively(path)`
+
+`cozy.client.files.createDirectory(path)` is used to create one or more directories for a given path.
```
enguerran: misnamed `createDirectory` instead of `createDirectoryRecursively`

enguerran:
If you miss time to cover code with tests, you may miss the point of *clean code*.

I don't think it is a good excuse.

I think a `TODO` is something that will never be done.

*My 2 cents with software craftsmanship inside*

gregorylegarec:
@enguerra I totally agree with you. I am just priorizing _clean sleep_ and _clean health_ over _clean code_.

gregorylegarec:
Ok I see no simple way to test this method easily, at last with the current mocking system based on MockAPI.

I tried to use sinon spies and stubs, without success.

PR #146, opened 2017-04-28
feat: expose cozyFetchJSON with cozy.client.fetchJSON()

gregorylegarec:


src/index.js
```diff
@@ -178,6 +179,13 @@ class Client {
     if (options.offline) {
       this.offline.init(options.offline)
     }
+
+    // Exposing cozyFetchJSON to make some development easier. Should be temporary.
+    this.fetchJSON = function _fetchJSON () {
```
gregorylegarec: Name function to make debugging easier (from @m4dz).

aenario: @gregorylegarec no arguments in arrow functions because you can instead use the ...spread attributes. As a bonus, this means you dont need to Array.prototype.slice.call

```javascript
this.fetchJSON = (...args) => cozyFetch.cozyFetchJSON.apply(this, [this].concat(args))
```

naming is cool thought!

PR #147, opened 2017-04-28
feat: Add skip and limit options to cozy.client.files.statByID

goldoraf:


src/files.js
```diff
@@ -143,18 +143,19 @@ export function trashById (cozy, id) {
   return cozyFetchJSON(cozy, 'DELETE', `/files/${encodeURIComponent(id)}`)
 }
 
-export function statById (cozy, id, offline = true) {
+export function statById (cozy, id, offline = true, options = null) {
```
gregorylegarec: What about `options = {}` ? To explicit the expected format and avoid `undefined variable options` when someone someday will do `if (options.foo)`

PR #155, opened 2017-05-10
Added wholeResponse param to cozy-data-query

goldoraf:
... and a small fix related to files sorting when offline

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -3,15 +3,10 @@ import { cozyFetch, cozyFetchJSON } from './fetch'
 import jsonapi from './jsonapi'
 import { DOCTYPE_FILES } from './doctypes'
 
-<<<<<<< HEAD
 // global variables
 export const ROOT_DIR_ID = 'io.cozy.files.root-dir'
 export const TRASH_DIR_ID = 'io.cozy.files.trash-dir'
 
-=======
-const ROOT_ID = 'io.cozy.files.root-dir'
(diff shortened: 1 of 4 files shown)


src/files.js
```diff
@@ -310,3 +310,10 @@ function toJsonApi (cozy, doc, contents = []) {
     }
   }
 }
+
+function sortFiles (allFiles) {
+  const folders = allFiles.filter(f => f.attributes.type === 'directory')
+  const files = allFiles.filter(f => f.attributes.type !== 'directory')
```
enguerran: Another way, with only one scan of the `allFiles` array:

```js
const {folders, files} = allFiles.reduce((obj, f) => {
    const type = f.attributes.type === 'directory' ? 'folders' : 'files'
    obj[type].push(f)
    return obj
}, {folders: [], files: []})
```

src/files.js
```diff
@@ -310,3 +310,10 @@ function toJsonApi (cozy, doc, contents = []) {
...
+  const sort = files => files.sort((a, b) => a.attributes.name.localeCompare(b.attributes.name))
+  return sort(folders).concat(sort(files))
+}
```
nono: You can probably do something shorter, like:

```js
function sortFiles (allFiles) {
  return allFiles.sort((a,b) => a.attributes.type.localCompare(b.attributes.type) || a.attributes.name.localeCompare(b.attributes.name))
}
```

PR #167, opened 2017-05-22
Add a jobs module

nono:
It includes works from #166


test/integration/jobs.js
```diff
@@ -0,0 +1,29 @@
...
+// eslint-disable-next-line no-unused-vars
+import should from 'should'
+import 'isomorphic-fetch'
+import {Client} from '../../src'
+
+const COZY_STACK_URL = process.env && process.env.COZY_STACK_URL || ''
+const COZY_STACK_VERSION = process.env && process.env.COZY_STACK_VERSION
+const COZY_STACK_TOKEN = process.env && process.env.COZY_STACK_TOKEN
+
+describe('jobs api', async function () {
```
gregorylegarec: Just a reminder for a detail, we should unify the way we describe our API in tests. Some are described as `settings` for example, other as `data API`.

PR #171, opened 2017-05-31
feat: add throw() method to intentService

gregorylegarec:


gregorylegarec:
@y-lohse That's @ptbrowne which asked for it. I think there can be ambiguous errors to handle: for example how to differenciate between a runtime error and a service-thrown error.

But let see how this can work.

test/unit/intents.js
```diff
@@ -543,6 +546,84 @@ describe('Intents', function () {
...
+          }
+
+          windowMock.parent.postMessage.callsFake(() => {
+            const messageEventListener = windowMock.addEventListener.secondCall.args[1]
+            messageEventListener(clientHandshakeEventMessageMock)
+          })
+
+          const service = await cozy.client.intents.createService(expectedIntent._id, windowMock)
+
+          service.throw(new Error('unique error'))
```
gregorylegarec: You're right, I will serialize it.

gregorylegarec:
@ptbrowne @y-lohse I added a serialization/deserialization for error, to pass one into postMessage(). What do you think about it ?

ptbrowne:
LGTM 👍

y-lohse:
I think this is good! My only concern is that this library is rapidly growing and at some point it may be easier to fall back to an open sourced library for this. Just a thing to keep in mind :)

gregorylegarec:
This debate, again. :)

PR #173, opened 2017-06-05
feat: create data with specific ID

enguerran:
A screenshot

When using ACH to import data in a cozy, we would like to import specific IDs in order to import relationship between objects.

To perform such a request, I suggest this code change. What do you think? Do you have another ~~ID~~ idea?

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -10,8 +10,10 @@ export function create (cozy, doctype, attributes) {
     if (isV2) {
       attributes.docType = doctype
     }
-    const path = createPath(cozy, isV2, doctype)
-    return cozyFetchJSON(cozy, 'POST', path, attributes).then((resp) => {
+    const path = createPath(cozy, isV2, doctype, attributes._id)
+    const httpVerb = attributes._id ? 'PUT' : 'POST'
+    delete attributes._id
+    return cozyFetchJSON(cozy, httpVerb, path, attributes).then((resp) => {
       if (isV2) {
         return find(cozy, doctype, resp._id)
       } else {


src/data.js
```diff
@@ -10,8 +10,10 @@ export function create (cozy, doctype, attributes) {
     if (isV2) {
       attributes.docType = doctype
     }
-    const path = createPath(cozy, isV2, doctype)
-    return cozyFetchJSON(cozy, 'POST', path, attributes).then((resp) => {
+    const path = createPath(cozy, isV2, doctype, attributes._id)
+    const VERB = attributes._id ? 'PUT' : 'POST'
```
nono: Why do you put `VERB` in upper case?

src/data.js
```diff
@@ -10,8 +10,10 @@ export function create (cozy, doctype, attributes) {
     if (isV2) {
       attributes.docType = doctype
     }
-    const path = createPath(cozy, isV2, doctype)
-    return cozyFetchJSON(cozy, 'POST', path, attributes).then((resp) => {
+    const path = createPath(cozy, isV2, doctype, attributes._id)
```
nono: Are you sure it works when there is no `_id`? It looks like `createPath` is expecting a blank string, not null. Maybe:

```js
const path = createPath(cozy, isV2, doctype, attributes._id || '')
```

src/data.js
enguerran: That was to reflect the value of verbs, but it may not be a very clever idea. ;-)

src/data.js
enguerran: If the `createPath`'s 4th parameter is `undefined`, it will be set to the [default value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) which is `''` (empty string).

Here is the `createPath` function: https://github.com/cozy/cozy-client-js/blob/master/src/utils.js#L62

src/data.js
enguerran: s/VERB/httpVerb/

PR #187, opened 2017-07-07
feat: Introduce cozy.client.data.findMany

sebn:


sebn:
Thanks for pointing this out @gregorylegarec, I'll check!

docs/data-api.md
```diff
@@ -43,6 +43,46 @@ console.log(doc._id, doc._rev, doc.title, doc.author, doc.isbn)
...
+documents found and identify the missing ones.
+
+- `doctype` is a string specifying the
+  [doctype](intro.md#doctypes--permissions)
+- `ids` is an array of strings specifying the identifiers of the documents you
+  look for
+
+**Warning**: Not available on **v2**. A fallback implementation could be
+provided at some point, but it would make one request per id anyway, since the
+old data system doesn't provide such an API.
```
nono: You are sure about that? `_all_docs` is needed for couchdb/pouchdb replication and we had that in cozy v2, so I'd have thought that the old data system can provide it.

sebn: @nono I looked at the [old data-system cookbook](https://docs.cozy.io/en/hack/cookbooks/data-system.html) (which doesn't seem to be available right now but still was 3 days ago). The cookbook didn't include anything about fetching multiple docs from a set of ids. But I didn't look at the data-system code, so no I'm not sure :)

I can remove the last part of the comment. Or do you think we should provide the v2 implementation right now?

sebn:
@gregorylegarec Looks like this can effectively happen when no document was created and the database doesn't exist yet. As soon as a document is created, even if it is immediately deleted, I assume it should not anymore.

- [ ] Handle missing database case

docs/data-api.md
nono: @sebn the documentation of the cozy-v2 API is far from complete. Just remove the last part of the comment

sebn: Done, thanks!

sebn:
I also made sure the resulting objects don't include useless `doc` / `error` properties (with `undefined` value).

:ship: as soon as CI is green.

gregorylegarec:
Thanks @sebn 👍

PR #189, opened 2017-07-19
Intents improvements (onReadyCallback, exposeFrameRemoval)

CPatchane:
Two main points:

* Handle `exposeIntentFrameRemoval` data flag to terminate the service without removing the intent DOM node directly but by providing the removal function to the client
__Example of usage__: Claudy has an animated (CSS) menu closing. Using this feature it will be possible to terminate the intent `.then` listen the animation ending before removing the intent iframe node using the resolved `removeIntentFrame ` function (and avoiding a sudden blank view in the Claudy view)

* Add an `onReadyCallback` optional argument to the `create` method in order to allow providing a callback function to be run when the intent iframe will be loaded (iframe `onload` listener).
__Example of usage__: Claudy will now wait the intent iframe to be totally loaded before beginning its opening animation in order to keep having a fluid UX.

__Edit:__ Rename `exposeRemoving` to `exposeFrameRemoval` and use a data flag instead of a method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - none yet
 
 ### Added
-- Add an exposeFrameRemoval method to the service to terminate him without removing the intent DOM node directly but by providing the function to do it to the client.
-- Add an onReadyCallback optional argument to the create method in order to allow providing a callback function to be run when the intent iframe will be loaded (iframe onload listener).
+- Handle `exposeIntentFrameRemoval` data flag to terminate the service without removing the intent DOM node directly but by providing the removal function to the client.
+- Add an `onReadyCallback` optional argument to the create method in order to allow providing a callback function to be run when the intent iframe will be loaded (iframe `onload` listener).
 
 ### Removed
 - none yet
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Removed
 - none yet
 
+## [v0.3.10] - 2017-07-XX
+### Changed
+- none yet
+
+### Fixed
+- none yet
+
+### Added
+- Add an exposeFrameRemoval method to the service to terminate him without removing the intent DOM node directly but by providing the function to do it to the client.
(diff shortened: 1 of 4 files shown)


y-lohse:
- `onReadyCallback` : no problem, pretty straightforward
- `exposeRemoving` : this one took me a while to figure out how it gets used, but it's ok now. I think `removeIntentFrame ` is a good name, but `exposeRemoving` not really. Maybe something like `exposeFrameRemoval` or `manuallyRemoveFrame` or whatever? `exposeRemoving` is not very clear I think.

So API-design wise that's all, the code is good too. I know you plan to update the changelog, but could you please also update the [intent documentation](https://github.com/cozy/cozy-client-js/blob/master/docs/intents-api.md)?

CPatchane:
Thanks @y-lohse! Sure, I was hesitating about the `exposeRemoving` name and your proposals sounds better so I will change that 👍 
And of course update the docs :)

y-lohse:
The changes are excellent! However I'm afraid I have a new question (just thought about it, sorry).
Right now, the *service* decides whether or not the iframe gets closed automatically, or if the client has to do it. This means clients must check for a `removeIntentFrame` property all the time, because they can't know if the `service` will do it or not. Is this correct? Have you considered doing it the other way around, ie. the *client* says "Please don't remove the frame when you're done"? Would that make sense?

CPatchane:
Hum, you're right, I'll think about that to be simpler and less error prone 👍

y-lohse:
Ok, cool. Sorry again I didn't mention it during the first review, the thought only occured to me later :/

CPatchane:
@y-lohse That was a very very good point in fact! Now it's even better, if the intent find a boolean `exposeIntentFrameRemoval` in its data provided by the client, the `terminate()` method will return the removal function with the doc autmatically, no need to use a different method in the service 🎸🎸 
See here https://github.com/cozy/cozy-client-js/pull/189/commits/3f56fb88f7c8e02c7fb51e4e8239ad4d405019bc

CPatchane:
Thanks @y-lohse 🎉

CPatchane:
[(5) [NUA-3] ETU mes applications n'ont pas besoin de permissions pour afficher Claudy](https://trello.com/c/gHTic2pp/501-5-nua-3-etu-mes-applications-nont-pas-besoin-de-permissions-pour-afficher-claudy)

PR #192, opened 2017-07-21
feat: option to allow bypassing credentials cache

y-lohse:
We have a case where the `scope` that is being requested might be different from the one that is cached, but the oauth flow always use the cached version, so the new scope is never picked up.

I've added a flag to the `client.authorize()` method that explicitly tells the oauth flow to ignore the cache. It defaults to false so existing calls shouldn't be affected. Does that make sense? Is there a better way to accomplish this?

### DIFF
--- a/src/auth_v3.js
+++ b/src/auth_v3.js
@@ -252,6 +252,10 @@ export function refreshToken (cozy, client, token) {
 // oauthFlow performs the stateful registration and access granting of an OAuth
 // client.
 export function oauthFlow (cozy, storage, clientParams, onRegistered, ignoreCachedCredentials = false) {
+  if (ignoreCachedCredentials) {
+    return storage.clear().then(() => oauthFlow(cozy, storage, clientParams, onRegistered, false))
+  }
+  
   let tryCount = 0
 
   function clearAndRetry (err) {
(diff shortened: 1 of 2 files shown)


y-lohse:
For cross-reference, this PR is needed by [this other PR on drive](https://github.com/cozy/cozy-drive/pull/334).

src/auth_v3.js
```diff
@@ -271,10 +271,14 @@ export function oauthFlow (cozy, storage, clientParams, onRegistered) {
...
-    storage.load(StateKey)
-  ])
+  const initialPromise = (ignoreCachedCredentials) ? storage.clear() : Promise.resolve()
+
+  return initialPromise.then(() => {
+    return Promise.all([
+      storage.load(CredsKey),
+      storage.load(StateKey)
+    ])
+  })
```
jinroh: I would have wrote it like that, for simplicity and to move this part out of the main promise chain:

```js
function oauthFlow (cozy, storage, clientParams, onRegistered, ignoreCachedCredentials = false) {
  if (ignoreCachedCredentials) {
    return storage.clear().then(() =>
      oauthFlow(cozy, storage, clientParams, onRegistered, false))
  }
  // ...
}
```

y-lohse: I agree, that makes it more readable. Thanks!

PR #194, opened 2017-08-07
Add intent transition handling

CPatchane:
Useful to 'animate' intent resizing using css transitions.

### DIFF
--- a/docs/intents-api.md
+++ b/docs/intents-api.md
@@ -48,12 +48,12 @@ If `intentId` and `window` parameters are not provided the method will try to re
 It returns a *service* object, which provides the following methods :
  * `getData()`: returns the data passed to the service by the client.
  * `getIntent()`: returns the intent
- * `resizeClient(doc)`: forces the size of the intent modale to a given width, maxWidth, height, maxHeight, or dimensions of a given element.
+ * `resizeClient(doc, transitionProperty)`: forces the size of the intent modale to a given width, maxWidth, height, maxHeight, or dimensions of a given element. The second optional argument `transitionProperty` can be used to add a CSS transition property on the intent in order to 'animate' the resizing.
  ```js
  // resize the client ot 300 pixels max height
  service.resizeClient({
     maxHeight: 300
- })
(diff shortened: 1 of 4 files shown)


y-lohse:
@CPatchane code looks good 👍 

Help me jog my memory though, when does this happen? When the service tells the client it wants to be displayed at a certain size, is that correct?

CPatchane:
Exactly this point @y-lohse

CPatchane:
[(5) [NUA-7] ETU je peux contacter le support directement depuis le nuage](https://trello.com/c/aGWhw9XW/541-5-nua-7-etu-je-peux-contacter-le-support-directement-depuis-le-nuage)

gregorylegarec:
I really don't get why an intent service should provide a presentation information to the client ? Isn't only client's purpose ?

CPatchane:
@gregorylegarec We could choose how the css is animated using CSS transition, couldn't we?

gregorylegarec:
@CPatchane Ok thanks, let's go with it 👍

y-lohse:
To be fair, this also bothers me a little. I think Claudy is an extreme use case of intents.
Theoretically speaking, would it be better if Claudy was distributed as a standalone js code, instead of an iframe? It could still use intents to do it's things, but only to manipulate data, while the resizing / animating / etc would be done on the client side and wouldn't event be in the intents?

(ping @CPatchane and @gregorylegarec so you get a notification just in case)

gregorylegarec:
The more it grows, the more Claudy become extreme in many way.


@y-lohse @CPatchane
Maybe the need to pass this kind of data from the intente service means that our whole design is bad. But also, having a standalone Claudy would not solve those issues.
However, externalizing Claudy from Cozy-Bar looks like a good idea. Again, we are falling back to our modularization and code mutualization problems.

CPatchane:
Agree with @gregorylegarec about separate Claudy would not solve issues about our current workflow.
However, I think we should see Claudy as a good example of extreme user of intents and a way to make it the most robust and reusable.

PR #205, opened 2017-08-30
Add one retry on index definition to avoid stack error

CPatchane:


src/mango.js
```diff
@@ -66,7 +66,21 @@ function defineIndexV3 (cozy, doctype, fields) {
...
+    .then((response) => {
+      const indexResult = { doctype: doctype, type: 'mango', name: response.id, fields }
+      const opts = getV3Options(indexResult, {'selector': {_id: {'$gt': null}}})
+      let path = createPath(cozy, false, indexResult.doctype, '_find')
+      return cozyFetchJSON(cozy, 'POST', path, opts)
+      .then(() => {
+        return indexResult
+      })
+      .catch(() => { // one retry
+        return cozyFetchJSON(cozy, 'POST', path, opts)
```
nono: I think you should wait a bit before retrying

CPatchane: Indeed, you're right 👍

gregorylegarec:
Thanks @CPatchane 🍰

y-lohse:
@CPatchane Can you give some background as to why we need to automatically do a `find` after the index is created please?

Without context it seems very strange to fire an unexpected request, whose result is discarded. It also breaks [here](https://github.com/cozy/cozy-client-js/pull/205/files#diff-7326a156fab6586d11c4a81a28c41bf9R71) if the `_id` field wasn't part of the index. Maybe we can find a better way to solve whatever the problem was 😃

CPatchane:
@y-lohse The problem was that at the first usage, when the index is created and returned to the client, it was actually not already available and cause an error on the client saying that the index doesn't exist. It's the case in production when using cozy-collect for the first time. I wanted to add some retries here to avoid this specific case. But I agree that I didn't test for other kind of index and I thought that all index was response on `_id` but I can be wrong.

y-lohse:
@CPatchane I can't reproduce the problem you're describing though :/. [Here's where I'm creating an index and using it right away](https://github.com/cozy/cozy-drive/pull/385/files#diff-a171b9a4b7e114f23187cf310589afd9R113). It's also surprising to me because we've been using indexes on photos for a while without problem.
Are you sure this wasn't an issue specific to cozy-collect? Maybe we could move this code to cozy-collect instead of having it here where it impacts every app using indexes, sometimes negatively when `_id` isn't part of the index?

CPatchane:
I know. This issue can happen only on production due to CouchDB usage (which is different from the local development) and only when you create the index for the very first time. It can be difficult to reproduce, but that could happen sometimes when a new user discover its new cozy. We reproduce it when we created new test intances on staging/production.

y-lohse:
Ugh, ok.
In that case, I suggest we do the following:

- check the `result` field in the response, which seems to be either `create` or `existing`, and only do the fetch on creation
- use one of the supplied fields in the selector instead of `_id`

Do you think this would work?

CPatchane:
@y-lohse Completely agree with your points here to improve that parts 👍 That should work for me.

y-lohse:
Cool, I'll probably submit a PR later today then :)

CPatchane:
Great @y-lohse Thanks 👍

PR #206, opened 2017-08-31
Improve data.findAll()

CPatchane:
- return directly array of docs
- doesn't include couchDB indexes

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -97,25 +97,13 @@ export function findAll (cozy, doctype) {
 
     return cozyFetchJSON(cozy, 'POST', path, {})
     .then((resp) => {
-      const result = {}
-      result.docs = []
+      const docs = []
 
       for (const row of resp.rows) {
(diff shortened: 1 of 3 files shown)


src/data.js
```diff
@@ -97,25 +97,13 @@ export function findAll (cozy, doctype) {
...
+      const docs = []
 
       for (const row of resp.rows) {
         const { doc } = row
-        result.docs.push(doc)
+        docs.push(doc)
       }
-      return result
-    })
-    .catch((error) => {
```
gregorylegarec: Why is the `catch` removed ?

CPatchane: Because, like a `.query()` using an index, we don't have to catch the error, it should be caught by the client directly. It was caught in `findMany` because we include errors in doc in the result object. Here we directly return the array.

gregorylegarec:
Thanks @CPatchane 👍

PR #209, opened 2017-09-18
feat: redirect to special reconnect page

y-lohse:
Adds a `reconnect` flag to the redirection URL so that a special login page is displayed.


y-lohse:
I have no idea why this test is failing as it has nothing to do with what I've changed in this PR..?

PR #213, opened 2017-10-04
doc: PouchDB needs to be attached to window in production

ptbrowne:
The production build of Bank was broken because of this.

In prod, since `cozy-client-js` is injected by the stack, it has not been processed by `wepback` and the `ProvidePlugin`. Thus you need to bind `PouchDB` to `window`.

### DIFF
--- a/docs/offline.md
+++ b/docs/offline.md
@@ -27,3 +27,13 @@ module.exports = {
 ```
 
 This same configuration has to be added in file `wepback.config.prod.js`, to make the build possible.
+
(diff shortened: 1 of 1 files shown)


ptbrowne:
Build is failing but I've only touched the doc :/

jinroh:
#214 should fix the failing tests.

ptbrowne:
> For future improvements, do you thinks it is possible to pass PouchDB and pouchdbFind as initialisation parameters to CozyClient ?

I think it would be a good idea to do so 👍, it would be an ersatz  of dependency injection. I think it would also be better from a testing point of view.

PR #214, opened 2017-10-04
Adapt jobs.count and add queued method

jinroh:
The `/jobs/queue` API has been changed in the stack to actually return the jobs data. This PR adapt `cozy-client-js` to this change.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -99,7 +99,8 @@ const intentsProto = {
 
 const jobsProto = {
   create: jobs.create,
-  count: jobs.count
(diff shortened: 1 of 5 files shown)


gregorylegarec:
@jinroh Do you have an idea why travis build are still failing ?

jinroh:
@gregorylegarec forgot to export the `queued` method on the main object. Although I don't really known yet where the second error is from.

gregorylegarec:
Thanks @jinroh 🍻

PR #215, opened 2017-10-05
Check iframe.parentNode existing when removing from DOM

CPatchane:


gregorylegarec:
Thanks @CPatchane :walking_woman:

PR #218, opened 2017-10-25
feat: improve iframe a11y

Narkfr:


src/intents.js
```diff
@@ -32,8 +32,10 @@ function injectService (url, element, intent, data, onReadyCallback) {
   // if callback provided for when iframe is loaded
   if (typeof onReadyCallback === 'function') iframe.onload = onReadyCallback
   iframe.setAttribute('src', url)
+  title && iframe.setAttribute('title', title)
   iframe.classList.add(intentClass)
   element.appendChild(iframe)
+  element.getElementsByClassName('coz-intent')[0].focus()
```
gregorylegarec: I think it's our iframe, so maybe we can just do:
```js
iframe.focus()
```
And avoid to query the DOM (Also, `cozy-intent` is already in a variable).

What do you think ?

src/intents.js
```diff
@@ -21,7 +21,7 @@ const errorSerializer = (() => {
 })()
 
 // inject iframe for service in given element
-function injectService (url, element, intent, data, onReadyCallback) {
+function injectService (url, element, intent, data, onReadyCallback, title) {
```
gregorylegarec: A few things here:
* To pass a `title` parameter here, we will need to pass it to the method `start` line 125.
* After `onReadyCallback`, the parameter list is beginning to grow. I suggest to start using on `options` parameter as last parameter. We need to keep the first ones `url`,  `element`,  `intent`, and `data` as they are all mandatory and important. So now using an `options` parameter now forces us to maintain legacy a little time. This `options` parameter should also be passed to the `start` method l. 125.
* The `title` parameter should not be mandatory, however it is a really nice thing to have one, if it improves accessibility. So we should set a default title.

I suggest to start this method this way now :
```js
const defaultTitle = `Cozy intent` // Should we avoid i18n for a default title ? 

[...]

function injectService  (url, element, intent, data, options = {}) {
  const isLegacyOnReadyCallback = typeof options === 'function'

  if (isLegacyOnReadyCallback) {
    console.warn && console.warn('onReadyCallback is now deprecated.')
  }

  const onReadyCallback = isLegacyOnreadyCallback ? options : options.onReadyCallback
  const title = options.title || defaultTitle
  [...]
```

What to you think ? (ping also @CPatchane and @y-lohse)

Narkfr:
[(8) [DRI-1] ETU je peux configurer un connecteur depuis Drive](https://trello.com/c/62xlnDeq/640-8-dri-1-etu-je-peux-configurer-un-connecteur-depuis-drive)

src/intents.js
Narkfr: I just add a 
`  // TODO: implement 'title' attribute` for now, because it's gonna take some time for me to add this feature, but I keep it in mind to do it any time in the near future.

gregorylegarec:
Thanks @Narkfr 👍

PR #231, opened 2017-11-27
feat: sanitize file name 🚑

gregorylegarec:
Having a trailing space in a Transifex translation, we ended up with a space in our file name, which was causing an error on Windows.


Narkfr:
LGTM

PR #238, opened 2017-12-01
Add babel polyfill dev peer dependency

ptbrowne:
- [x] we need babel-polyfill for tests to run
- [x] tell developers using cozy-client-js that they need babel-polyfill


ptbrowne:
I'm betting as a cache in Travis as the yarn.lock did not change.

PR #240, opened 2017-12-06
Doc: add doc about intent size

Narkfr:


CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Removed
 - none yet
 
+## [v0.3.21] - 2017-11-02
+### Addes
```
ptbrowne: added

ptbrowne:
I would put the doc closer to `resizeClient` method definition since here as a reader, I don't even know what resizeClient is, what an intent is etc...

I would also put an example of where it is appropriate to use resizeClient : claudy ? I am not knowledgeable on this. ping @CPatchane

ptbrowne:
LGTM. We can merge but I still do not know what is `resizeClient` good for.

kosssi:
Comment c'est possible que ça soit passé ? :0

PR #242, opened 2018-01-29
feat: add support for heic and heif files

y-lohse:
On older devices, the OS doesn't compute a `type` for `HEIC` and `HEIF` files, so I added a manual check as a fallback option.

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -13,6 +13,12 @@ function sanitizeFileName (name) {
   return name && name.trim()
 }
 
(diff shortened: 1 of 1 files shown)


y-lohse:
[(2) 📱 Envoyer le bon mime-type pour les images HEIC](https://trello.com/c/5FpicYAR/784-2-%F0%9F%93%B1-envoyer-le-bon-mime-type-pour-les-images-heic)

PR #249, opened 2018-02-16
Allow to pass Content-Length when uploading file

aenario:


src/files.js
```diff
@@ -61,14 +61,17 @@ function doUpload (cozy, data, method, path, options) {
     lastModifiedDate = new Date(lastModifiedDate)
   }
 
+  const headers = {
+    'Content-Type': contentType
+  }
+  if (contentLength) headers['Content-Length'] = String(contentLength)
+  if (checksum) headers['Content-MD5'] = checksum
+  if (lastModifiedDate) headers['Date'] = lastModifiedDate.toGMTString()
+  if (ifMatch) headers['If-Match'] = ifMatch
```
ptbrowne: Have you noticed a problem when a header was set to '' ?

test/integration/files.js
```diff
@@ -13,6 +13,7 @@ const COZY_STACK_VERSION = process.env && process.env.COZY_STACK_VERSION
 const COZY_STACK_TOKEN = process.env && process.env.COZY_STACK_TOKEN
 
 describe('files API', async function () {
+  this.timeout(5000)
```
ptbrowne: TIL

test/integration/files.js
```diff
@@ -59,6 +60,47 @@ describe('files API', async function () {
...
+    stream.push('datastring1')
+    stream.push(null)
+
+    const created = await cozy.client.files.create(stream, {
+      name: filename,
+      contentType: 'application/json',
+      contentLength: 4
+    })
+
+    // created.should.be.rejectedWith(/content size/)
```
ptbrowne: why is it commented ?

test/integration/files.js
```diff
@@ -59,6 +60,47 @@ describe('files API', async function () {
...
+    stream.push('datastring1')
+    stream.push(null)
+    // stream.timeout = 2000
+
+    const oldFetch = global.fetch
+    global.fetch = (url, opts) => {
+      if (opts) opts.timeout = 500
+      return oldFetch(url, opts)
+    }
+    after(() => { global.fetch = oldFetch })
```
ptbrowne: I did not know about this after() function and I cannot see it in the [Jest doc](https://facebook.github.io/jest/docs/en/api.html). Where does it come from ?

ptbrowne: OK, we use mocha here.

test/integration/files.js
aenario: This was the expected behaviour but it actually ends up as a timeout

src/files.js
aenario: I had an issue with an undefined header wich was sent as "undefined", I feel this better express the need.

test/integration/files.js
aenario: Correction : The expected behaviour was that we got an error from the stack, but somewhere in the node-fetch / http.Request mechanism, the request actually get truncated.

src/files.js
ptbrowne: 👍

test/integration/files.js
aenario: removed commented code

PR #253, opened 2018-03-05
Ability to filter services returned by the stack

ptbrowne:
If an application wants a specific service for its intent,
it can specify `filterServices` in the option when creating
the intent.


gregorylegarec:
Maybe the `filterServices` should need a little documentation in the API, particulary to describe the service object which is an internal mechanism of the intents.

More the intents are evolving, more the `data` parameter is used as `options` (at first it was only used to pass document related informations). Let's keep this in mind for future improvements.

PR #254, opened 2018-03-16
feat: add notification properties in auth v3

drazik:
Add properties needed by the stack to handle push notification with oauth client. See [cozy-stack doc](https://cozy.github.io/cozy-stack/auth.html#post-authregister) for more infos.


drazik:
Here it is

Issue #201, opened 2017-08-28
Evolving intents


y-lohse:
Lately, we have discovered new use cases for intents, and in the past we've had to add  things to the Intent API that weren't part of the initial plan. We should start thinking about the next version of the intent specs.

Before we start, two strongly recomended reads:
- [The first version of the spec](https://github.com/cozy/cozy-stack/blob/master/docs/intents.md)
- [The actual API documentation](https://github.com/cozy/cozy-client-js/blob/master/docs/intents-api.md#intents)

## A) Things that are not in the spec but are in the API

1) A hook for the client to know when the iframe is loaded — See #189, used to wait until the service is loaded to display it.
1) A hook for the client to remove the iframe manually — See #189, used to animate the iframe closing.
1) The client suggest a size for the service — See #175, but I'm not sure I understand the use case very well.
1) The service tells the client to change the size of the service's frame — See #175 and #194, used when a service wants to be displayed at a specific size and the client couldn't predict it. Can happen after the service is initialized or when the user interacts with the service (eg. expanding Claudy).

### 1. Linking to a specific part of an app

The most common example for this is wanting to link to a certain document inside cozy-drive. Except cozy-drive, no one has any clue what the exact URL is — in fact, even the URL for the drive app is mostly unknown for apps.

We've discussed this during a meeting, and agreed that the best solution is to introduce the notion of [disposition](https://github.com/cozy/cozy-stack/blob/master/docs/intents.md#disposition). The signature would be something like this: `cozy.client.intents.create('OPEN', 'io.cozy.files', {id: '123abc'}, null, 'window')` (note the `window` at the end, indicating the disposition).
Once started, the service can then change the browser's URL (or tell the client to do it).

### 2. ???

You tell me. I'll try to keep this post updated.

y-lohse:
Now a few subjective points.

There's *probably* a new use case, which is "Silent Computing". A client wants another app to perform some computation on some data (probably data the client has not access to) and get a result back, but the service shouldn't be visible.
This can be hacked by hiding the service iframe, but maybe it should become a legit thing. It also needs very special consideration, because while intents in general are there to work around permissions, it's usually with the user's consent. In this case... maybe not ☠️ .
I think @Ljinod and @Gara64 are interested in this, and @ptbrowne may have actually done it. I may also have a use case for this, but not sure yet.

[...]

PR #256, opened 2018-04-02
feat: add intents.redirect() method ✨

gregorylegarec:
~PRIORITY : VERY LOW~

This PR tries to solve the redirection point exposed in #201 and adds two methods to intents API, both used to easily redirect to a given document or a given doctype. The two added methods are:

* `getRedirectionURL(type, data)` which retrieves a redirection URL.
* `redirect(type, data)` which retrieves a redirection URL and performs the redirection.

### Example

```jsx
import Button from 'cozy-ui/react/Button'
// [...]
  render() {
    const { folder } = this.props
    return (
      <Button
        label={`Open ${folder.name}`}
        onClick={() => cozy.client.intents.redirect('io.cozy.files', folder)}
      />
    )
  }
```

### Declaring a redirection handler

Redirection are declared in `manfiest.webapp`, like a regular intent

```json
"intents": [
    {
      "action": "REDIRECT",
      "type": [
        "io.cozy.konnectors"
      ],
      "href": "/"
    }
  ],
```

### Current limitations
* For now, it is still not possible to remove an intent from database.
* Statck cannot handle an intent having a sharp character in this `href` attribute, and URLencode it (i.e. replaces `#` by `%23`).

### DIFF
--- a/src/intents.js
+++ b/src/intents.js
@@ -238,3 +238,43 @@ export function createService (cozy, intentId, serviceWindow) {
         })
     })
 }
+
+// Redirect to an app able to handle the doctype
+// Redirections are more or less a hack of the intent API to retrieve an URL for
+// accessing a given doctype or a given document.
+// It needs to use a special action `REDIRECT`
+export async function getRedirectionURL (cozy, type, data) {
+  if (!type && !data) throw new Error(`Cannot retrieve redirection, at least type or doc must be provided`)
+
+  const intent = await create(cozy, 'REDIRECT', type, data)
+
+  const service = pickService(intent)
+  if (!service) throw new Error('Unable to find a service')
+
+  // Intents cannot be deleted now
+  // await deleteIntent(cozy, intent)
+
+  // ignore query string and intent id
+  const baseURL = service.href.split('?')[0]
+  // FIXME: Handle the fact that the stack encode the '#' character in the URL
+  const sanitizedURL = baseURL.replace('%23', '#')
+  return data ? buildRedirectionURL(sanitizedURL, data) : sanitizedURL
+}
+
+function isSerializable (value) {
+  return !['object', 'function'].includes(typeof value)
+}
+
+function buildRedirectionURL (url, data) {
+  const parameterStrings = Object.keys(data)
+    .filter(key => isSerializable(data[key]))
+    .map(key => `${key}=${data[key]}`)
+
+  return parameterStrings.length ? `${url}?${parameterStrings.join('&')}` : url
+}
+
+export async function redirect (cozy, type, doc) {
+  if (!window) throw new Error('redirect() method can only be called in a browser')
+  const redirectionURL = await getRedirectionURL(cozy, type, doc)
+  window.location.href = redirectionURL
+}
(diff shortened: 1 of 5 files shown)


src/intents.js
```diff
@@ -108,6 +108,16 @@ function injectService (url, element, intent, data, onReadyCallback) {
 }
 
 const first = arr => arr && arr[0]
+// In a far future, the user will have to pick the desired service from a list.
+// For now it's our job, an easy job as we aribtrary pick the first service of
```
CPatchane: typo s/aribtrary/arbitrary

src/intents.js
```diff
@@ -108,6 +108,16 @@ function injectService (url, element, intent, data, onReadyCallback) {
...
+// the list.
+function pickService (intent, filterServices) {
```
CPatchane: 👍

CPatchane:
For the `#` in the URL, maybe we can add an exception? Same thing for the `?` and `&` characters since they can be used in a URL? Just a suggestion, maybe it's not necessary for now.

src/intents.js
y-lohse: Isn't that comment out of date though ? The first service is the default choice now, but the `filterServices` already allows you to pick one, right?

src/intents.js
```diff
@@ -234,3 +238,40 @@ export function createService (cozy, intentId, serviceWindow) {
...
+  const baseURL = service.href.split('?')[0]
+  return doc ? parseRedirectionURL(baseURL, doc) : baseURL
+}
+
+function parseRedirectionURL (url, doc) {
+  // replace /:params/ segments
+  return url.replace(/:([^//]+)/g, segment => {
+    const param = segment.substring(1)
+    if (!doc[param]) throw new Error(`Cannot generate redirection URL, document misses ${param} attribute.`)
+    return doc[param]
```
y-lohse: Neat 👍

src/intents.js
ptbrowne: s/aribtrary/arbitrarily/

src/intents.js
gregorylegarec: As the `filterServices` is a callback given as parameter, it's not the user which pick the service, the choice is still made by code.

gregorylegarec:
@CPatchane Actually, it's the stack wich generate the intent `href` and replace `#` by `%23` into it. I am not sure where it can be changed.

CPatchane:
I think we could merge the two actions to use only a `REDIRECT` action and then pass parameters to be handled in the targeted application.

gregorylegarec:
After meeting with @CPatchane, we agreed on improving the current implementation by using only a `REDIRECT` action and adding parameters in query string. This implies a better and explicit redirection handling in target application, but is more flexible and allow more possibilities.

The `manifest.webapp` declaration could evolve to enumerate the allowed parameters, first as information to contributor and developer, and maybe in the future for an automated use by the stack.

Example:

```json
"intents": [
    {
      "action": "REDIRECT",
      "type": [
        "io.cozy.konnectors"
      ],
      "href": "/#/discover",
      "params": ["slug, "category", "dataType"]
    }
  ],
```

gregorylegarec:
[(3) ETU je peux filtrer dans mon Store](https://trello.com/c/HK3JrzHB/917-3-etu-je-peux-filtrer-dans-mon-store)

CPatchane:
A fix has been added to handle the `#` character but must be removed when it will be correctly handled by the stack (ping @nono)

PR #257, opened 2018-04-11
feat: asynchronous errors in intents.createService() ✨

gregorylegarec:
First errors in `intents.createService()` were throwing error synchronously. To be catch by `.catch()` method, they must be thrown with a Promise rejection.


gregorylegarec:
@y-lohse Yep, we should rewrite this part with `async/await` when we will move it to the future `cozy-interapp` library.

PR #259, opened 2018-04-16
Use `babel-preset-cozy-app`

CPatchane:
Remove custom config to use cozy-app, it fixes the regenerator polyfill issue.


gregorylegarec:
Thanks @CPatchane 📦

PR #263, opened 2018-04-23
chore: add Prettier

gregorylegarec:


CPatchane:
I hope to make optional react usage in `eslint-config-cozy-app` in order to use it in a simpler way for JS only libraries, but waiting that LGTM 😀

PR #264, opened 2018-04-26
enhancement: add composition feature on intents

gregorylegarec:
This PR adds intent composition feature. It allows, from an intent service, to request another intent. As this intent have to be handled by the intent client, due to `X-Frame-Options` limitations, it has been named _Intent Composition_.

An intent composition can be called this way:
```js
const service = cozy.client.intents.createService()
// ...
const app = service.compose('INSTALL', 'io.cozy.apps', { slug: 'drive' })
```
What happens internally:
* The service send a `compose` message to the client
* The client handle this message, start a new intent and hides the current one
* Once the intent is resolve, it passes the result to the first intent, via a message to the service
* The service is resolved.

This process can be chained indefinitely.

### Refactoring
I managed to refactor a little bit the code. What I had in mind was more modular and more clear than the actual result. I am not satisfied by the client part. But the [code related to the composition](https://github.com/cozy/cozy-client-js/pull/264/commits/ad47bb65faa2f22f08e6f7b64e479c6236811086)  is actually relatively simple and clear. I plan to improve the whole intent code base in the future.


gregorylegarec:
[(5) [COL] ETU je peux installer/configurer UN connecteur depuis de l'inter-apps [3]](https://trello.com/c/PnUUxk1F/898-5-col-etu-je-peux-installer-configurer-un-connecteur-depuis-de-linter-apps-3)

src/intents/service.js
```diff
@@ -47,17 +47,39 @@ export function start(cozy, intentId, serviceWindow) {
   return cozyFetchJSON(cozy, 'GET', `/intents/${intentId}`).then(intent => {
     let terminated = false
 
-    const terminate = message => {
+    const sendMessage = message => {
```
CPatchane: 👍

CPatchane:
Thanks @gregorylegarec 🏒

PR #268, opened 2018-06-08
chore: declare iso fetch as dependency

y-lohse:
`isomorphic-fetch` is an actual dependency for node, so it can't be in `devDependency` or it won't get installed.


enguerran:
In order to take into account the commit messages styleguide proposed here https://github.com/cozy/cozy-guidelines/pull/46 I reword your commit.

For next time, it will be mandatory you follow these guidelines.

y-lohse:
How dare you (thanks)

PR #270, opened 2018-06-20
chore: upgrade pouch

y-lohse:
I'd like to upgrade the PouchDB dependencies here, mainly because pouchdb-find on its current version has abyssal performances (at least compared to the newer version).

If I'm not mistaken, this change shouldn't have any effect outside of Drive, since I think it's the only app that uses the offline module. Am I wrong?


y-lohse:
@nono it seems that there are indeed 2 versions of leveldown, one comes from pouch, the other from [cozy-data-system](https://github.com/cozy/cozy-data-system) which I had never heard about 😮. 

Still not sure how to resolve that, and why there wouldn't be *any* bindings at all…

nono:
Cozy-data-system is for Cozy v2 (nodejs). I think we can remove Cozy v2 stuff from this repository. The other options are:

- Throw away yarn and come back to npm. Yarn is doing stupid here, it should use only one version (1.5.0) of leveldown that matches the dependency for both pouch and the cozy-data-system.
- Fix a specific version of leveldown (1.5.0) in package.json.

nono:
> and why there wouldn't be any bindings at all…

I think that when there are several versions of the same module, yarn use a different layout on the file system. The JS files are still found, but not the bindings.

y-lohse:
[(5) [forum-5712] Le tri est lent](https://trello.com/c/UxXUjiar/965-5-forum-5712-le-tri-est-lent)

PR #272, opened 2018-06-22
feat(intents): add an redirectFn parameter to redirect method

drazik:
This adds a `redirectFn` so we can pass a custom redirection function when we call `cozy.client.intents.redirect`.

This enables us to use an inapp browser and redirect the user using intents redirect on mobile apps, for example.


gregorylegarec:
Thanks @drazik 👍

nono:
I'm a bit late, but I'm not sure to see the value of adding a `redirectFn` paramater to `redirect` vs calling `getRedirectionURL` and doing whatever you want with the returned URL.

PR #274, opened 2018-07-30
Properly compute intents base URL with hash

CPatchane:
The hash part is now provided at the end of the `href` by the stack, ccjs was skipping it then. Now it's handled properly.

### DIFF
--- a/src/intents/index.js
+++ b/src/intents/index.js
@@ -42,6 +42,10 @@ export function createService(cozy, intentId, serviceWindow) {
   return service.start(cozy, intentId, serviceWindow)
 }
 
+function removeQueryString(url) {
+  return url.replace(/\?[^/#]*/, '')
+}
+
 // Redirect to an app able to handle the doctype
 // Redirections are more or less a hack of the intent API to retrieve an URL for
 // accessing a given doctype or a given document.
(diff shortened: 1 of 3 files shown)


CPatchane:
[Lorsque je clique sur 'Ajouter un compte' dans Banks, je suis redirigé vers le store mais ce dernier n'est pas filtré sur les services bancaires](https://trello.com/c/uj5hvIsq/1136-lorsque-je-clique-sur-ajouter-un-compte-dans-banks-je-suis-redirig%C3%A9-vers-le-store-mais-ce-dernier-nest-pas-filtr%C3%A9-sur-les-servic)

src/intents/index.js
```diff
@@ -61,7 +61,7 @@ export async function getRedirectionURL(cozy, type, data) {
   // await deleteIntent(cozy, intent)
 
   // ignore query string and intent id
-  const baseURL = service.href.split('?')[0]
+  const baseURL = service.href.replace(/\?[\w-=&]*/, '')
```
ptbrowne: I do not understand what this is supposed to do :/ I think naming the regex should help.

Can you also add test cases ?

CPatchane: By requesting an intent from the stack, the stack will provide an intent ID like `http://store.cozy.tools:8080/?intent=85ef0631669a4ea34ecf9ca6d10298d5#/redirect`. Previously, the hash part was before the query and the previous code worked. But now the hash part is provided at the end, it's why we use this regex to remove the query part only.

Good point. I updated the tests, thanks 👍

PR #277, opened 2018-09-17
Migrate away from pouch-websql

y-lohse:
I added a function that allows us to migrate databases that are using websql to pouchdb — see https://pouchdb.com/2018/01/23/pouchdb-6.4.2.html

I was experiencing a weird build error that I had seen before, where, webpack was injecting code that was referencing a `global` variable, but that variable was undefined and the code broke. This was an old bug in webpack, so I updated us to the first version that worked, which is v3.


y-lohse:
[(3) 📱🐛 Erreur WebSQL sous iOS](https://trello.com/c/fZhVxxLK/954-3-%F0%9F%93%B1%F0%9F%90%9B-erreur-websql-sous-ios)

PR #279, opened 2018-10-30
Expose fetch method

gregorylegarec:
It is now convenient to expose this method to fetch icons from stack.

Used in applications which did not switch to cozy-client yet.


gregorylegarec:
@ptbrowne I updated code with `options` parameter. 👌

PR #283, opened 2019-03-18
Upgrade pouchdb

### LINKED ISSUE #282: Upgrading pouchdb
This package currently uses pouchdb v6, and we should upgrade to pouch v7. v6 is not maintained and has some vulnerabilities.

The main change is that the websql driver is not packaged by default anymore. We shouldn't need it — the only known use case for it is a page running in an iframe on safari mobile, so an intent for example. But no app does this at the moment.  
That said, the websql driver can be added as a separate dependency, and that would probably be the safest way to upgrade.

When we do this, we should also switch to the browser-only version of pouch — it's significantly lighter.

doubleface:
Here is the goal:

- avoid including leveldown in nodejs version of cozy-client-js -> pouchdb-browser
- upgrade pouchdb to 7 to avoid security problems
- remove dependencies to cozy v2 libs

Fixes #282


Crash--:
> It upgrades pouchdb to v7, where the websql is a separate plugin. I think we still use websql in interapp/iframes on drive. Am I wrong ?

Yep. We removed it few weeks ago https://github.com/cozy/cozy-drive/commit/a9b2bb83d7dc5e9d0522ecea5e405139dea524b8 

> It removes any local pouchdb like the cordova sqlite plugin. Isn't Drive still on cozy-client-js ? If so, does this PR breaks drive and its storage ?

ATM, we don't use sqlite. And plan to use it soon, but we can do it in drive.

PR #285, opened 2019-03-20
Transpile instead of bundling

ptbrowne:
- [x] Files are transpiled
- [ ] Babel-preset-cozy-app should be upgraded and tuned with
presetEnv: {modules: false}, transformRuntime: { helpers: false }.
- [ ] Need to be shipped with tag beta on npm and tested


nono:
I don't know if it is still used, but the cozy-stack serves a bundled version of cozy-client-js: https://github.com/cozy/cozy-stack/blob/master/assets/.externals#L13-L19

ptbrowne:
OK. Here I am mainly targetting node JS, it should help with konnector size.

nono:
Do we even need transpilation? I don't think we are using anything new that is not supported by Node 8 in this repository (but maybe we have an older version of nodejs for konnectors, I don't remember).

ptbrowne:
Good question. We need transpilation since import is not yet fully supported in node.js. It is behind the flag `--experimental-modules` and the extension of your files must be `.mjs`.

nono:
For konnectors, we are doing a bundle to have a single .js file to execute. So, I don't think it would be an issue for konnectors (but we need to transform imports for cozy-desktop).

ptbrowne:
ACH is using `cozy-client-js` without bundling. So it would be an issue if `import`s were used.

ptbrowne:
Closing as cozy-client-js is no longer used except by Desktop.

PR #286, opened 2019-03-27
fix: include isomorphic-fetch in node build

doubleface:
isomorphic-fetch will allow pouchdb-browser to not cause 'ReferenceError: fetch is not defined' error


y-lohse:
Do we have some background on this error? Did `pouchdb-browser` just crash out of the box in node?
The plan was that the `pouchdb` part of the code never runs in node, so `fetch` shouldn't be required..?

doubleface:
> Do we have some background on this error? Did `pouchdb-browser` just crash out of the box in node?
> The plan was that the `pouchdb` part of the code never runs in node, so `fetch` shouldn't be required..?

This fetch error appears when the pouchdb-browser package is required, because it is supposed to be used only in the browser and there is no reason for fetch to not be present.

Crash--:
@y-lohse Is it what you have in mind : https://github.com/cozy/cozy-client-js/compare/fix/import?expand=1

y-lohse:
> @y-lohse Is it what you have in mind : https://github.com/cozy/cozy-client-js/compare/fix/import?expand=1

I wasn't aware that simply requiring pouchdb would produce an error like that to be honest. But ok, since it does, I think not loading it in a node context is a better long term solution than exposing a fetch function. Not blocking for me though, depends how urgent the fix is.

nono:
Even without pouchdb, fetch is used in a lot of places in cozy-client-js and isomorphic fetch is listed as a dependency for using cozy-client-js with node:

> If you are developing a nodejs app, you will need some polyfills. You can add them to your project with this command:
> $ yarn add isomorphic-fetch core-js regenerator-runtime btoa

https://github.com/cozy/cozy-client-js/blob/master/docs/intro.md#include-the-library-in-your-application

@doubleface can you update this bit of documentation too?

PR #288, opened 2019-05-24
Add the possibility to use cursor-based pagination for references

Gara64:
The recommended way to paginate view results in [CouchDB](https://docs.couchdb.org/en/stable/ddocs/views/pagination.html#paging-alternate-method), and therefore the stack, is through [cursors](https://docs.cozy.io/en/cozy-stack/jsonapi/#pagination).
This is because `skip` performs badly on large data collections, as it requires to scan the full B-Tree. 

For instance, a query on 2500 documents, with a limit of 100 docs per request, was taking more than 25s overall, with sometimes more than 1s per paginated request.

In `fetchReferencedFiles`, the sort was forced on `datetime`, which, behind the hoods, triggers the use of the `referenced-by-sorted-by-datetime` view (see [here](https://github.com/cozy/cozy-stack/blob/4e8da5cc1e5401ea43029de6d90bf248fac82405/web/files/references.go#L77-L81 ) and [here](https://github.com/cozy/cozy-stack/blob/ba71942b33a006d6b6a6bdfa39626292f27be5c5/pkg/couchdb/index.go#L71-L84))
This view uses  `[doctype, id, datetime]` keys, which is useful to get all the references sorted on the datetime through partial match queries on the key, e.g. `key=[io.cozy.photos.albums, "album-id", {}]`. However, it is not possible to use partial-match queries with cursor-based pagination as it requires the use of a `startkey_docid` field, which makes only sense when there are multiple results for a single key, not multiple ones.

So, this PR allows to use an `id` sort, which queries the `referenced-by` view, with `[doctype, id]` keys. Therefore, one can query a particular album with `key=[io.cozy.photos.albums, "album-id"]` and paginate the results with cursors through `startkey_docid`.

### DIFF
--- a/docs/files-api.md
+++ b/docs/files-api.md
@@ -336,28 +336,46 @@ const href = await cozy.client.files.getArchiveLinkByIds(["1592673"], "secretpro
 ### `cozy.client.data.addReferencedFiles(doc, fileIds)`
 
 `cozy.client.data.addReferencedFiles(doc, fileIds)` binds the files to the document.
-(see cozy-stack [documentation](https://cozy.github.io/cozy-stack/references-docs-in-vfs.html) for more details)
+(see cozy-stack [documentation](https://docs.cozy.io/en/cozy-stack/references-docs-in-vfs/) for more details)
 
 
 ### `cozy.client.data.removeReferencedFiles(doc, fileIds)`
 
 `cozy.client.data.removeReferencedFiles(doc, fileIds)` unbinds the files to the document.
-(see cozy-stack [documentation](https://cozy.github.io/cozy-stack/references-docs-in-vfs.html) for more details)
+(see cozy-stack [documentation](https://docs.cozy.io/en/cozy-stack/references-docs-in-vfs/) for more details)
 
 
 ### `cozy.client.data.listReferencedFiles(doc)`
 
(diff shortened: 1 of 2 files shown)


docs/files-api.md
```diff
@@ -336,29 +336,47 @@ const href = await cozy.client.files.getArchiveLinkByIds(["1592673"], "secretpro
...
 
 It returns a promise for a list of files.
 
+There are 2 alternatives for pagination, that can be used through the `option` parameter: `cursor` and `skip`. For more details, see how the stack handles [pagination](https://docs.cozy.io/en/cozy-stack/jsonapi/#pagination). Note that cursor-based pagination performs way better than skip-based on large volumes.
+
+* `option` is an object with the following fields:
+  * `cursor` (recommended): specify the view's key and the starting docid. The starting docid can be empty for the first query and take the last returned docid for the next ones.
+  * `skip` (not recommended): ignore the first x results for pagination.
+  * `limit`: maximum number of results.
+* `sort` is a string which can be `datetime` (default) or `id`.
```
ptbrowne: Can you add the warning "Sort=datetime will not work with cursor based pagination" ?

nono:
@Gara64 can you rebase master? I've made a PR to fix the build on travis (#289)

PR #292, opened 2019-06-12
feat: allow to define any file metadata on file upload

doubleface:
This follows api defined in https://github.com/cozy/cozy-stack/pull/1929.
Metadata is created in io.cozy.files.metadata, and metadata id is referenced in file creation options.


src/files.js
```diff
@@ -74,7 +80,13 @@ function doUpload(cozy, data, method, path, options) {
   if (lastModifiedDate) headers['Date'] = lastModifiedDate.toGMTString()
   if (ifMatch) headers['If-Match'] = ifMatch
 
-  return cozyFetch(cozy, path, {
+  let finalpath = path
+  if (metadata) {
+    const metadataId = await sendMetadata(cozy, metadata)
+    if (metadataId) finalpath += `&MetadataID=${metadataId}`
```
nono: Are we sure that the path always contains a query-string?

doubleface: You are right, if doUpload is called by `updateById`, there is no query-string. I will fix that

doubleface: Fixed

doubleface:
@nono I did the fixes after your comments. Are you ok now ?

src/files.js
```diff
@@ -2,6 +2,8 @@
 import { cozyFetch, cozyFetchJSON } from './fetch'
 import jsonapi from './jsonapi'
 import { DOCTYPE_FILES } from './doctypes'
+import qs from 'querystring'
```
nono: Does it work in a browser? Or does it mean that cozy-client-js will only be used in cozy-desktop and the konnectors?

edas: querystring should work in browsers : https://www.npmjs.com/package/querystring

Browser support is not optional, cozy-client-js is still used by frontend apps

src/files.js
```diff
@@ -90,6 +106,20 @@ function doUpload(cozy, data, method, path, options) {
   })
 }
 
+function addQueryString(path, querystring = {}) {
+  const pathurl = url.parse(path)
+  const query = qs.decode(pathurl.query)
```
edas: style (for lines 111-112):

```suggestion
  const query = { ...qs.decode(pathurl.query), ...querystring }
```

src/files.js
```diff
@@ -90,6 +106,20 @@ function doUpload(cozy, data, method, path, options) {
...
+  Object.assign(query, querystring)
+  return pathurl.pathname + '?' + qs.encode(query)
+}
+
+async function sendMetadata(cozy, metadata) {
+  const result = await cozyFetchJSON(cozy, 'POST', '/files/upload/metadata', {
+    data: { type: 'io.cozy.files.metadata', attributes: metadata }
+  })
+  return result && result._id ? result._id : false
```
edas: style (if prettier let us), because such priorities are not clear at first sight

```suggestion
  return (result && result._id) ? result._id : false
```

src/files.js
nono: But, querystring must be in package.json for that to work, no?

edas: Yes it should, and it doesn't. Goog catch

src/files.js
taratatach: This can also be written like so:
```suggestion
  return (result && result._id) || false
```

src/files.js
doubleface: I think cozy-drive still uses it for some tasks. Then I will do it another way

src/files.js
doubleface: Prettier does not let us, sorry

src/files.js
```diff
@@ -2,6 +2,8 @@
 import { cozyFetch, cozyFetchJSON } from './fetch'
 import jsonapi from './jsonapi'
 import { DOCTYPE_FILES } from './doctypes'
+import qs from 'querystring'
+import url from 'url'
```
y-lohse: This is also node only AFAIK.

src/files.js
```diff
@@ -74,7 +82,15 @@ function doUpload(cozy, data, method, path, options) {
   if (lastModifiedDate) headers['Date'] = lastModifiedDate.toGMTString()
   if (ifMatch) headers['If-Match'] = ifMatch
 
-  return cozyFetch(cozy, path, {
+  let finalpath = path
```
y-lohse: Why are we introducing a new `finalPath` variable?

doubleface:
@edas @nono I remove the use of querystring which is a node internal module and I saw that cozy-drive uses cozy-client-js for some features.

I tested it with a connector in 'int' environment an I can se the metadata in io.cozy.files.

src/files.js
doubleface: I introduced it because I don't like to change function parameters.

edas:
Also, shouldn't we put /dist in gitignore ?

src/files.js
edas: I second that. Changing function parameters is not really a recommended pattern

doubleface:
@edas I was not part of this decision to keep the dist directory in git. But since now travis builds this directory itself before publishing, I suppose we could remove it.

PR #293, opened 2019-06-14
chore: ignore dist directory

doubleface:
Since now travis builds cozy-client-js itself before publishing, there is no need to keep the dist directory git.
This makes it a lot easier to foller diffs and PRs.


Crash--:
👍

PR #295, opened 2019-06-19
feat: allow to add sourceAccount and sourceAccountIdentifer in cozyMetadata while uploading a file

doubleface:
sourceAccount is the io.cozy.accounts account id associated to the connector which uploaded the file
sourceAccountIdentifier is the user identifier (login most of the time) associated to the account

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -46,7 +46,9 @@ async function doUpload(cozy, data, method, path, options) {
     checksum,
     lastModifiedDate,
     ifMatch,
-    metadata
+    metadata,
+    sourceAccount,
+    sourceAccountIdentifier
   } = options || {}
   if (!contentType) {
     if (isBuffer) {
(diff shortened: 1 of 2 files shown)


src/files.js
```diff
@@ -90,6 +92,18 @@ async function doUpload(cozy, data, method, path, options) {
...
+  if (sourceAccount) {
+    finalpath += `${
+      finalpath.includes('?') ? '&' : '?'
+    }SourceAccount=${sourceAccount}`
+  }
+
+  if (sourceAccountIdentifier) {
+    finalpath += `${
+      finalpath.includes('?') ? '&' : '?'
+    }SourceAccountIdentifier=${sourceAccountIdentifier}`
```
nono: Maybe you can extract a small helper function that adds a parameter to the query-string in `finalpath`. It will avoid to repeat those 3 lines several times.

doubleface: I considered it but found it more readable like this finally. Since the options and the query string parameters are not exactly the same, we have to pass to many parameters.

But maybe you are right. With a good naming, it may be better.

PR #301, opened 2019-10-11
docs: Add a deprecated notice in the TOC

Crash--:
You want that cozy-client-js not being indexed
in our official documentation since is a bit
confusing for our users.

If really needed, the user can still access
to the doc on github

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -1,3 +1,3 @@
-⚠️ cozy-client-js is a depreacted librairy. We recommend you to use our new client : [Cozy Client](https://docs.cozy.io/en/cozy-client/getting-started/)
+⚠️ cozy-client-js is a deprecated library. We recommend to use our new client : [Cozy Client](https://docs.cozy.io/en/cozy-client/getting-started/)
 
 If you really need to see the documention of cozy-client-js, see it on [github](https://github.com/cozy/cozy-client-js/blob/master/docs/README.md)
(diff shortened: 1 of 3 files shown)


docs/deprecated.md
```diff
@@ -0,0 +1,3 @@
+⚠️ cozy-client-js is a depreacted librairy. We recommend you to use our new client : [Cozy Client](https://docs.cozy.io/en/cozy-client/getting-started/)
```
ptbrowne:
```suggestion
⚠️ cozy-client-js is a deprecated library. We recommend to use our new client : [Cozy Client](https://docs.cozy.io/en/cozy-client/getting-started/)
```

docs/deprecated.md
```diff
@@ -0,0 +1,3 @@
...
+
+If you really need to see the documention of cozy-client-js, see it on [github](https://github.com/cozy/cozy-client-js/blob/master/docs/README.md)
```
ptbrowne:
```suggestion
If you really need to see the documentation of cozy-client-js, see it on [github](https://github.com/cozy/cozy-client-js/blob/master/docs/README.md)
```

docs/deprecated.md
nono: [Nitpick] No space before `:` in english

docs/deprecated.md
nono: Maybe we can set up again github pages for cozy-client-js

ptbrowne: IMHO, we should not spend more time on cozy-client-js and the github documentation is sufficient. Do you see any advantage of github pages over normal github for documentation (apart from being nicer looking) ?

nono: It looks nicer, and it is just two clicks to setup.

PR #302, opened 2020-01-17
feat: add method to fetch capabilities

taratatach:
The Cozy's capabilities route is separated from the rest of the
instance's settings so we need a method to fetch this resource.
It's necessary to know which Swift layout is used by the Cozy and if
it's using flat or nested subdomains.

### DIFF
--- a/src/settings.js
+++ b/src/settings.js
@@ -19,6 +19,10 @@ export function updateInstance(cozy, instance) {
   return cozyFetchJSON(cozy, 'PUT', `/settings/instance`, instance)
 }
 
+export function getCapabilities(cozy) {
+  return cozyFetchJSON(cozy, 'GET', `/settings/capabilities`)
+}
+
 export function getClients(cozy) {
   return cozyFetchJSON(cozy, 'GET', `/settings/clients`)
 }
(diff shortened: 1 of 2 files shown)


src/settings.js
```diff
@@ -19,6 +19,10 @@ export function updateInstance(cozy, instance) {
   return cozyFetchJSON(cozy, 'PUT', `/settings/instance`, instance)
 }
 
+export function getCapabilities(cozy) {
+  return cozyFetchJSON(cozy, 'GET', `/settings/capabilities`)
+}
```
ptbrowne: I think that we should add features to cozy-client instead of cozy-client-js. Instantiating a cozy-client from a cozy-client-js client is fairly easy and would pave the way for the complete migration to cozy-client.

ptbrowne: I am uncomfortable to have this here where cozy-client does not have this feature.

ptbrowne: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L185

Crash--: We have it in cozy-client! SettingCollection.get('capabilities')

Crash--: >Instantiating a cozy-client from a cozy-client-js client is fairly easy and would pave the way for the complete migration to cozy-client.

I told that to @taratatach a few weeks ago. We should really push in that direction.

taratatach:
This is not necessary anymore and Cozy Desktop now uses `cozy-client` when necessary so if we ever need to access the capabilities, we'll use `cozy-client` as well.
Thanks for the review though :).

PR #310, opened 2021-02-11
feat: Acknowledge `noSanitize` in all file methods

taratatach:
By default document names are sanitized (i.e. trimmed) when making
requests to the Cozy.
However, some clients like Cozy Desktop require that the names they
pass to `cozy-client-js` are passed on untouched to the Cozy (in case
of Cozy Desktop, doing otherwise will lead to a de-synchronized state
between the local and remote trees).

We already have a `noSanitize` option that prevents the sanitation
from happening but it's only available when creating files via the
`create` method.
We now add and use this option in all methods that would otherwise
sanitize the document name by default.


LucsT (approved):
Ok pour moi. 
Ne devrait pas changer le comportement par défault.

doubleface (approved):

Crash--:
We should have added at least one test in https://github.com/cozy/cozy-client-js/blob/master/test/unit/files.js

PR #325, opened 2022-07-27
chore: Use Couch 3.2.2

Crash--:


.travis.yml
```diff
@@ -13,7 +13,7 @@ cache:
   directories:
     - node_modules
 before_script:
-  - docker run -d -p 5984:5984 --name couch apache/couchdb:2.3
+  - docker run -d -p 5984:5984 --name couch -e COUCHDB_USER= -e COUCHDB_PASSWORD= apache/couchdb:3.2.2
```
taratatach, line 16: Are you sure we can use empty admin credentials? I believe this is not possible and would recommend using `admin/admin` or something similar.

Crash--: Just trying to debug, so trying several things ^^. 

> Are you sure we can use empty admin credentials? I believe this is not possible and would recommend using admin/admin or something similar.

Yeah, but I'll have to edit the cozy.yaml file before running the stack. If I understand correctly, actually we don't have this step. I was trying to avoid it.

